PR #18, opened 2020-02-16
Use different mount point for each PV

tonicmuroq:
Made some changes:
* use `/jfs/$(volumeID)` as mount root, `volumeID` is different for each PVC, so each mount process can use its own mount path, then multiple versions will be supported.
* clean both mount paths and mount ref paths, make sure node driver is cleaned when PVC is deleted.

### DIFF
--- a/pkg/driver/controller.go
+++ b/pkg/driver/controller.go
@@ -59,6 +59,7 @@ func (d *controllerService) CreateVolume(ctx context.Context, req *csi.CreateVol
 		volCtx[k] = v
 	}
 	// subPath is not supported in dynamic provisioning
+	klog.V(5).Infof("CreateVolume Warning: `subPath` is not supported in dynamic provisioning")
 	delete(volCtx, "subPath")
 
 	volume := csi.Volume{
@@ -71,7 +72,7 @@ func (d *controllerService) CreateVolume(ctx context.Context, req *csi.CreateVol
 
 // DeleteVolume moves directory for the volume to trash (TODO)
 func (d *controllerService) DeleteVolume(ctx context.Context, req *csi.DeleteVolumeRequest) (*csi.DeleteVolumeResponse, error) {
-	// klog.V(4).Infof("DeleteVolume: called with args: %#v", req)
+	klog.V(4).Infof("DeleteVolume: called with args: %#v", req)
 	volumeID := req.GetVolumeId()
 
 	if len(volumeID) == 0 {
--- a/pkg/driver/controller.go
+++ b/pkg/driver/controller.go
@@ -56,11 +56,11 @@ func (d *controllerService) CreateVolume(ctx context.Context, req *csi.CreateVol
 		return nil, status.Error(codes.InvalidArgument, "Volume Capabilities cannot be empty")
 	}
 
-	capReuired := req.CapacityRange.GetRequiredBytes()
-	if capa, ok := d.vols[req.Name]; ok && capa < capReuired {
-		return nil, status.Errorf(codes.AlreadyExists, "Volume: %q, capacity bytes: %d", req.Name, capReuired)
+	requiredCap := req.CapacityRange.GetRequiredBytes()
+	if capa, ok := d.vols[req.Name]; ok && capa < requiredCap {
+		return nil, status.Errorf(codes.AlreadyExists, "Volume: %q, capacity bytes: %d", req.Name, requiredCap)
 	}
-	d.vols[req.Name] = capReuired
+	d.vols[req.Name] = requiredCap
 
 	volCtx := make(map[string]string)
 	for k, v := range req.Parameters {
@@ -70,7 +70,7 @@ func (d *controllerService) CreateVolume(ctx context.Context, req *csi.CreateVol
 
 	volume := csi.Volume{
 		VolumeId:      req.Name,
-		CapacityBytes: capReuired,
+		CapacityBytes: requiredCap,
 		VolumeContext: volCtx,
 	}
 	return &csi.CreateVolumeResponse{Volume: &volume}, nil
--- a/pkg/driver/controller.go
+++ b/pkg/driver/controller.go
@@ -43,10 +43,7 @@ func newControllerService() controllerService {
 }
 
 // CreateVolume create directory in an existing JuiceFS filesystem
-func (d *controllerService) CreateVolume(
-	ctx context.Context,
-	req *csi.CreateVolumeRequest) (
-	*csi.CreateVolumeResponse, error) {
+func (d *controllerService) CreateVolume(ctx context.Context, req *csi.CreateVolumeRequest) (*csi.CreateVolumeResponse, error) {
 	// DEBUG only, secrets exposed in args
 	// klog.V(5).Infof("CreateVolume: called with args: %#v", req)
 
@@ -57,38 +54,24 @@ func (d *controllerService) CreateVolume(
 		return nil, status.Error(codes.InvalidArgument, "Volume Capabilities cannot be empty")
 	}
 
-	secrets := req.Secrets
-	klog.V(5).Infof("CreateVolume: Secrets contains keys %+v", reflect.ValueOf(secrets).MapKeys())
-
-	capacityBytes := int64(req.CapacityRange.GetRequiredBytes())
-
-	klog.V(5).Infof("CreateVolume: Mounting juicefs %q", req.Name)
-	jfs, err := d.juicefs.JfsMount(secrets, []string{})
-	if err != nil {
-		return nil, status.Errorf(codes.Internal, "Could not mount juicefs: %v", err)
+	volCtx := map[string]string{"mode": "dynamic"}
+	for k, v := range req.Parameters {
+		volCtx[k] = v
 	}
-
-	klog.V(5).Infof("CreateVolume: Creating volume %q, capacity bytes: %d", req.Name, capacityBytes)
-	vol, err := jfs.CreateVol(req.Name, capacityBytes)
-	if status.Convert(err).Code() == codes.AlreadyExists {
-		return nil, err
-	}
-
-	if err != nil {
-		return nil, status.Errorf(codes.InvalidArgument, "Could not create volume %q size %d bytes: %v", req.Name, capacityBytes, err)
-	}
+	// subPath is not supported in dynamic provisioning
+	delete(volCtx, "subPath")
 
 	volume := csi.Volume{
 		VolumeId:      req.Name,
-		CapacityBytes: vol.CapacityBytes,
-		VolumeContext: req.Parameters,
+		CapacityBytes: req.CapacityRange.GetRequiredBytes(),
+		VolumeContext: volCtx,
 	}
 	return &csi.CreateVolumeResponse{Volume: &volume}, nil
 }
 
 // DeleteVolume moves directory for the volume to trash (TODO)
 func (d *controllerService) DeleteVolume(ctx context.Context, req *csi.DeleteVolumeRequest) (*csi.DeleteVolumeResponse, error) {
-	klog.V(4).Infof("DeleteVolume: called with args: %#v", req)
+	// klog.V(4).Infof("DeleteVolume: called with args: %#v", req)
 	volumeID := req.GetVolumeId()
 
 	if len(volumeID) == 0 {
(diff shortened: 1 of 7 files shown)


pkg/driver/controller.go
```diff
@@ -63,7 +63,7 @@ func (d *controllerService) CreateVolume(
 	capacityBytes := int64(req.CapacityRange.GetRequiredBytes())
 
 	klog.V(5).Infof("CreateVolume: Mounting juicefs %q", req.Name)
-	jfs, err := d.juicefs.JfsMount(secrets, []string{})
+	jfs, err := d.juicefs.JfsMount(req.Name, secrets, []string{})
```
yujunz: Maybe we should defer "umount" here. The mount point is not needed once the `CreateVol` is done.

pkg/juicefs/juicefs.go
```diff
@@ -177,14 +179,14 @@ func NewJfsProvider(mounter *mount.SafeFormatAndMount) (Interface, error) {
 }
 
 // JfsMount auths and mounts juicefs
-func (j *juicefs) JfsMount(secrets map[string]string, options []string) (Jfs, error) {
+func (j *juicefs) JfsMount(volumeID string, secrets map[string]string, options []string) (Jfs, error) {
 	stdoutStderr, err := j.AuthFs(secrets)
```
yujunz, line 110: juicefs client is used by `AuthFs` as well. Better to upgrade before this.

pkg/driver/controller.go
```diff
@@ -98,7 +98,7 @@ func (d *controllerService) DeleteVolume(ctx context.Context, req *csi.DeleteVol
 	secrets := req.Secrets
 	klog.V(5).Infof("DeleteVolume: Secrets contains keys %+v", reflect.ValueOf(secrets).MapKeys())
 
-	jfs, err := d.juicefs.JfsMount(secrets, []string{})
+	jfs, err := d.juicefs.JfsMount(volumeID, secrets, []string{})
```
yujunz, line 91: Better to defer `Unmount` here instead of inside `DeleteVol`.

pkg/juicefs/juicefs.go
```diff
@@ -161,7 +161,9 @@ func (fs *jfs) DeleteVol(volName string) error {
 	}
 	stdoutStderr, err := jfsProvider.RmrDir(path.Join(fs.MountPath, volName))
 	klog.V(5).Infof("DeleteVol: rmr output is '%s'\n", stdoutStderr)
-	return err
+
+	klog.V(5).Infof("DeleteVol: unmount %s", fs.MountPath)
+	return jfsProvider.Unmount(fs.MountPath)
```
yujunz: It might be better defer this Unmount right after Mount as commented above.

yujunz (changes requested):

pkg/juicefs/juicefs.go
yujunz: I think `version --upgrade` upgrades `juicefs.py` itself too, does it? @davies

tonicmuroq: `juicefs version --upgrade` will upgrade `juicefs` too

pkg/driver/controller.go
tonicmuroq: right, controller driver will be very clean too

yujunz:
It seems travis-ci is asking for go 1.13. Not sure why it didn't fail before.

```
note: module requires Go 1.13FAIL	github.com/juicedata/juicefs-csi-driver/tests/sanity [build failed]
333M
```

Feel free to modify the configuration to make it pass. @tonicbupt

tonicmuroq:
a little change in `JFS_AUTO_UPGRADE`:
* if it's not set, then auto upgrade is ignored
* if it's an integer value, the value will be timeout in seconds for upgrading
* if it's not an integer, then timeout will be set to `10` seconds as default

tonicmuroq:
deploy/k8s.yaml not changed yet, i accidentally clicked `ready for review` button =___=

yujunz:
> a little change in `JFS_AUTO_UPGRADE`:
> 
> * if it's not set, then auto upgrade is ignored
> * if it's an integer value, the value will be timeout in seconds for upgrading
> * if it's not an integer, then timeout will be set to `10` seconds as default

This is too tricky. Better to name JFS_UPGRADE_TIMEOUT for readability.

tonicmuroq:
@yujunz as you said... after i removed code from `Controller.CreateVolume` and sanity tests failed... 
* now two env vars will be used: if `JFS_AUTO_UPGRADE` is not set, jfsmount will not upgrade automatically. if `JFS_AUTO_UPGRADE` is set and `JFS_AUTO_UPGRADE_TIMEOUT` is set to an integer, this value will be used as upgrade timeout, and default to 10s.
* `CreateVol` is now lazily called in `NodePushlishVolume`
* Dynamic provisioning will add `mode` in `VolumeContext`, and will be used in `NodePushlishVolume` to identify dynamic / static provisioning

since I have tested all the yamls in example directory, is it possible to remove sanity tests, otherwise `CreateVol` is still necessary in controller driver

pkg/driver/controller.go
```diff
@@ -57,38 +58,20 @@ func (d *controllerService) CreateVolume(
...
-
 	volume := csi.Volume{
 		VolumeId:      req.Name,
-		CapacityBytes: vol.CapacityBytes,
-		VolumeContext: req.Parameters,
+		CapacityBytes: req.CapacityRange.GetRequiredBytes(),
+		VolumeContext: map[string]string{
+			"mode":     "dynamic",
+			"capacity": strconv.FormatInt(req.CapacityRange.GetRequiredBytes(), 10),
+		},
```
tonicmuroq: And there's no way to get the name of required PVC from PVC ID...
or k8s address will be needed here

pkg/driver/node.go
```diff
@@ -118,37 +119,62 @@ func (d *nodeService) NodePublishVolume(ctx context.Context, req *csi.NodePublis
...
+				capacity = v
+			}
+		}
+
+		vol, err := jfs.CreateVol(volumeID, capacity)
+		if err != nil {
+			return nil, status.Errorf(codes.Internal, "Could not create volume: %s, %v", volumeID, err)
+		}
+
+		bindSource = vol.VolumePath
```
tonicmuroq: dynamic provisioning will use `VolumePath` which is a temporary path

pkg/driver/node.go
```diff
@@ -118,37 +119,62 @@ func (d *nodeService) NodePublishVolume(ctx context.Context, req *csi.NodePublis
...
+	} else {
+		// static provisioning
+		bindSource = jfs.GetBasePath()
```
tonicmuroq: for static provisioning, `GetBasePath()` returns the root of JFS

pkg/driver/node.go
```diff
@@ -118,37 +119,62 @@ func (d *nodeService) NodePublishVolume(ctx context.Context, req *csi.NodePublis
...
+	}
 	if subPath, ok := req.GetVolumeContext()["subPath"]; ok {
 		bindSource = path.Join(bindSource, subPath)
```
tonicmuroq: if pre-existing directory is needed, `subPath` will be used, relative to root path

pkg/driver/node.go
```diff
@@ -118,37 +119,62 @@ func (d *nodeService) NodePublishVolume(ctx context.Context, req *csi.NodePublis
...
-	klog.V(4).Infof("NodeUnpublishVolume: called with args %+v", req)
+	// klog.V(4).Infof("NodeUnpublishVolume: called with args %+v", req)
 
 	target := req.GetTargetPath()
 	if len(target) == 0 {
 		return nil, status.Error(codes.InvalidArgument, "Target path not provided")
 	}
 
+	klog.V(5).Infof("NodeUnpublishVolume: unmounting ref for target %s", target)
+	if meta, err := juicefs.ReadMetaFile(target); err == nil {
```
tonicmuroq: mount point and volume path are stored in meta file. this seems to be the only way to get mount point from `target`, since we bind a sub directory inside JFS to `target`, `GetMountRefs` can not find the correct mount point, so meta file is needed here, to cleanup JFS root mount point. otherwise there's gonna be a leak of jfsmount processes.

tonicmuroq:
@davies @yujunz

yujunz:
> * `CreateVol` is now lazily called in `NodePushlishVolume`

This does not  feel right, might be violating the spec. See https://github.com/container-storage-interface/spec/blob/master/spec.md#createvolume 

The return code `CreateVolume` indicates the result of provisioning the storage, so we need to check auth and quota (when supported in the future).

yujunz:
Besides, node driver may not have the permission to create volume in a restrict configuration.

tonicmuroq:
it's soooo ugly to cheat sanity ... @yujunz @davies

tonicmuroq:
ok ... @davies @yujunz

pkg/driver/controller.go
```diff
@@ -57,38 +54,24 @@ func (d *controllerService) CreateVolume(
...
-	}
-
-	if err != nil {
-		return nil, status.Errorf(codes.InvalidArgument, "Could not create volume %q size %d bytes: %v", req.Name, capacityBytes, err)
+	volCtx := map[string]string{"mode": "dynamic"}
+	for k, v := range req.Parameters {
+		volCtx[k] = v
 	}
+	// subPath is not supported in dynamic provisioning
+	delete(volCtx, "subPath")
```
yujunz: Left a warning log here?

pkg/driver/controller.go
```diff
@@ -57,38 +54,24 @@ func (d *controllerService) CreateVolume(
...
+		CapacityBytes: req.CapacityRange.GetRequiredBytes(),
+		VolumeContext: volCtx,
 	}
 	return &csi.CreateVolumeResponse{Volume: &volume}, nil
 }
 
 // DeleteVolume moves directory for the volume to trash (TODO)
 func (d *controllerService) DeleteVolume(ctx context.Context, req *csi.DeleteVolumeRequest) (*csi.DeleteVolumeResponse, error) {
-	klog.V(4).Infof("DeleteVolume: called with args: %#v", req)
+	// klog.V(4).Infof("DeleteVolume: called with args: %#v", req)
```
yujunz: Why removing the log?

pkg/driver/node.go
```diff
@@ -118,40 +119,67 @@ func (d *nodeService) NodePublishVolume(ctx context.Context, req *csi.NodePublis
...
-	bindSource := jfs.GetBasePath()
-	if subPath, ok := req.GetVolumeContext()["subPath"]; ok {
-		bindSource = path.Join(bindSource, subPath)
+	var (
+		subPath string
+		volRoot string
+	)
+	if m, ok := volCtx["mode"]; ok && m == "dynamic" {
+		// dynamic provisioning
+		// /jfs/$(subPath)/$(volumeID) if subPath
```
yujunz: This is inconsistent with

```
	// subPath is not supported in dynamic provisioning
 	delete(volCtx, "subPath")
```

yujunz: If `subPath` is not supported by dynamic provisioning, we can simply assign `volumeID` to `subPath` to make the logic consistent. If `subPath` is supported, concat `subPath` and `volumeID` to make the directory for dynamic provisioning, and `subPath` only when static. This would allow you to substitute `CreateVol` with `Mkdir`.

What do you think?

davies: +1

tonicmuroq: i was just leaving code here in case we'll support `subPath` in the future (by creating volume in controller driver)...

tonicmuroq: since `subPath` is not supported in dynamic provisioning, i removed `volRoot` parameter in `CreateVol`, now `subPath` means the relative path from JFS root.

pkg/driver/node.go
```diff
@@ -118,22 +119,27 @@ func (d *nodeService) NodePublishVolume(ctx context.Context, req *csi.NodePublis
...
 	}
 
-	bindSource := jfs.GetBasePath()
-	if subPath, ok := req.GetVolumeContext()["subPath"]; ok {
-		bindSource = path.Join(bindSource, subPath)
+	subPath := ""
+	if sp, ok := volCtx["subPath"]; ok {
+		subPath = sp
+	}
+	bindSource, err := jfs.CreateVol(volumeID, subPath)
```
davies: subPath -> volCtx["subPath"]

pkg/driver/controller.go
```diff
@@ -86,7 +70,7 @@ func (d *controllerService) CreateVolume(
 
 	volume := csi.Volume{
 		VolumeId:      req.Name,
-		CapacityBytes: vol.CapacityBytes,
+		CapacityBytes: capReuired,
```
yujunz:
```suggestion
		CapacityBytes: capRequired,
```

yujunz (commented):
Typo

davies (approved):

tonicmuroq:
works fine for all examples @yujunz @davies

yujunz (approved):

PR #22, opened 2020-02-19
fix pvc temporary dir

tonicmuroq:
add `volumeID` as `subPath` and pass all the parameters to node publish phase


pkg/driver/controller.go
```diff
@@ -78,10 +78,15 @@ func (d *controllerService) CreateVolume(
 		return nil, status.Errorf(codes.InvalidArgument, "Could not create volume %q size %d bytes: %v", req.Name, capacityBytes, err)
 	}
 
+	volCtx := map[string]string{"subPath": req.Name}
+	for k, v := range req.Parameters {
+		volCtx[k] = v
```
davies, line 83: Add a check for k != "subPath"

tonicmuroq: override `subPath` later

PR #31, opened 2020-03-02
no cronjob for pinned version

tonicmuroq:
this can be merged when next version is released

### DIFF
--- a/.github/workflows/release-pinned.yaml
+++ b/.github/workflows/release-pinned.yaml
@@ -1,11 +1,9 @@
-name: release
+name: release-pinned
 
 on:
   push:
     branches:
       - master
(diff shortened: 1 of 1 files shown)


.github/workflows/release-pinned.yaml
```diff
@@ -0,0 +1,41 @@
+name: release-pinned
+
+on:
+  push:
+    branches:
+      - master
```
tonicmuroq, line 6: may be we should use `tag` to trigger this...

tonicmuroq:
see https://github.com/juicedata/juicefs-csi-driver/pull/34 and https://github.com/juicedata/juicefs-csi-driver/pull/33

PR #43, opened 2020-03-23
Fix dynamic-provisioning example Kustomization

camper42:
Fix `Error: json: unknown field "env"`


yujunz:
This is due to the version of kustomize. Which version are you using? @camper42

camper42:
`{Version:3.5.4 GitCommit:3af514fa9f85430f0c1557c4a0291e62112ab026 BuildDate:2020-01-17T14:23:25+00:00 GoOs:darwin GoArch:amd64}`

copy from static-provisioning
https://github.com/juicedata/juicefs-csi-driver/blob/master/examples/static-provisioning/kustomization.yaml

yujunz:
That explains. The existing version is for kustomize 2.x which is the embedded version in `kubectl -k`.

Need to have a way to deal with such compatibility issue.  cc @tonicbupt

yujunz:
Could you please update README along with your patch? @camper42 

Maybe leave a note that kustomize 3.x is required.

PR #44, opened 2020-03-23
mount options should be passed to `mount`

tonicmuroq:
`juicefs` will separate options bebind `-o` to fuse options and mount options, e.g. `-o cache-size=100,cache-dir=/var/foo,allow-root` to `--cache-size=100 --cache-dir=/var/foo` as mount options and `--allow-root` as fuse options. hence they should be joined by `','`, not separated as a list.
and `options` were not used in `Mount`, didn't know why...

### DIFF
--- a/pkg/driver/node.go
+++ b/pkg/driver/node.go
@@ -18,10 +18,8 @@ package driver
 
 import (
 	"context"
-	"fmt"
 	"os"
 	"reflect"
-	"strings"
 
 	csi "github.com/container-storage-interface/spec/lib/go/csi"
 	"github.com/juicedata/juicefs-csi-driver/pkg/juicefs"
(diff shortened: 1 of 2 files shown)


pkg/driver/node.go
```diff
@@ -115,13 +113,7 @@ func (d *nodeService) NodePublishVolume(ctx context.Context, req *csi.NodePublis
...
 	mountOptions := []string{}
 	if opts, ok := volCtx["mountOptions"]; ok {
-		mountOptions = strings.Split(opts, ",")
-	}
-	for k, v := range options {
-		if v != "" {
-			k = fmt.Sprintf("%s=%s", k, v)
-		}
-		mountOptions = append(mountOptions, k)
+		mountOptions = append(mountOptions, opts)
```
yujunz: This is changing the logic. Has it been verified that mountOptions accepts.a comma separated string?

Before the change, it was accepting a string array.

yujunz (commented):
The options was omitted during AuthFs but it should be applied in MountFs

pkg/driver/node.go
tonicmuroq: juicefs accepts `-o opt1=val1,opt2=val2,opt3=val3` because juicefs use `,` to separate string followed by `-o`, if `mountOptions` is an array like `["opt1=val1", "opt2=val2", "opt3=val3"]`, then juicefs will receive `-o opt1=val1 opt2=val2 opt3=val3`, the latter two are ignored. but I need to check `j.Mount` to see how `options` is passed first...

tonicmuroq:
```
func makeMountArgs(source, target, fstype string, options []string) []string {
    // Build mount command as follows:
    //   mount [-t $fstype] [-o $options] [$source] $target
    mountArgs := []string{}
    if len(fstype) > 0 {
        mountArgs = append(mountArgs, "-t", fstype)
    }
    if len(options) > 0 {
        mountArgs = append(mountArgs, "-o", strings.Join(options, ","))
    }
    if len(source) > 0 {
        mountArgs = append(mountArgs, source)
    }
    mountArgs = append(mountArgs, target)

    return mountArgs
}
```

oops, it will join again, so there shouldn't be a bug here... I'll check more...

PR #47, opened 2020-03-28
Use alpine as base image

yujunz:
* Image size reduced from 173M to 45M
* Use `juicefs mount` instead of `mount` to Handle error `no Python available, please install it first` under alpine

### DIFF
--- a/pkg/juicefs/juicefs.go
+++ b/pkg/juicefs/juicefs.go
@@ -18,7 +18,6 @@ import (
 const (
 	cliPath   = "/usr/bin/juicefs"
 	mountBase = "/jfs"
-	fsType    = "juicefs"
 	// DefaultCapacityBytes is 10 Pi
 	DefaultCapacityBytes = 10 * 1024 * 1024 * 1024 * 1024 * 1024
 )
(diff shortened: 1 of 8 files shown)


pkg/juicefs/juicefs.go
```diff
@@ -204,9 +203,11 @@ func (j *juicefs) MountFs(volumeID, name string, options []string) (string, erro
 		return mountPath, status.Errorf(codes.Internal, "Could not check mount point %q exists: %v", mountPath, err)
 	}
 
+	args := append([]string{"mount", name, mountPath}, options...)
```
tonicmuroq: `j.Mount` will add `-o` before `options`, so `"-o"` need to be in `args`

tonicmuroq: `args := append([]string{"mount", name, mountPath, "-o"}, options...)` @yujunz 

`options` are like `cache-dir=/var/foo,cache-size=100`

yujunz: OK. Will correct and test it before merging

tonicmuroq (approved):

PR #55, opened 2020-07-15
add support for --no-update option for AuthFs

chnliyong:
The `juicefs.py` now has a `--no-update` option. When sub command `auth` and `mount` with this option, it would only use local config(not update config from `CFG_URL`). The `mount` sub command can use `mountOptions` to support this.

This PR use `secret` to transfer initial local config looks a little bit strange...

### DIFF
--- a/pkg/juicefs/juicefs.go
+++ b/pkg/juicefs/juicefs.go
@@ -206,7 +206,8 @@ func (j *juicefs) AuthFs(secrets map[string]string) ([]byte, error) {
 			if _, err := os.Stat(confPath); os.IsNotExist(err) {
 				err = ioutil.WriteFile(confPath, []byte(secrets["initconfig"]), 0644)
 				if err != nil {
-					klog.V(5).Infof("Create config file: %q failed", confPath)
+					return nil, status.Errorf(codes.Internal,
+						"Create config file %q failed: %v", confPath, err)
 				}
 				klog.V(5).Infof("Create config file: %q success", confPath)
 			}
--- a/pkg/juicefs/juicefs.go
+++ b/pkg/juicefs/juicefs.go
@@ -3,6 +3,7 @@ package juicefs
 import (
 	"context"
 	"fmt"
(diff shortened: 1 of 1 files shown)


pkg/juicefs/juicefs.go
```diff
@@ -191,6 +192,26 @@ func (j *juicefs) AuthFs(secrets map[string]string) ([]byte, error) {
...
+		}
+		if secrets["initconfig"] != "" {
+			conf := secrets["name"] + ".conf"
+			confPath := filepath.Join("/root/.juicefs", conf)
+			if _, err := os.Stat(confPath); os.IsNotExist(err) {
+				err = ioutil.WriteFile(confPath, []byte(secrets["initconfig"]), 0644)
+				if err != nil {
+					klog.V(5).Infof("Create config file: %q failed", confPath)
+				}
+				klog.V(5).Infof("Create config file: %q success", confPath)
```
davies, line 212: this line should be in  logged when `err == nil`

pkg/juicefs/juicefs.go
```diff
@@ -191,6 +192,26 @@ func (j *juicefs) AuthFs(secrets map[string]string) ([]byte, error) {
 			argsStripped = append(argsStripped, fmt.Sprintf("--%s=[secret]", k))
 		}
 	}
+	if v, ok := os.LookupEnv("JFS_NO_UPDATE_CONFIG"); ok && v == "enabled" {
+		args = append(args, "--no-update")
+		argsStripped = append(argsStripped, "--no-update")
+
+		if secrets["bucket"] == "" {
```
davies, line 199: bucket could be already in `initconfig`

chnliyong: Currently `juicefs auth` not use `bucket` from local config, if we want the default value of `bucket` from local config, we shoud change `juicefs.py`

yujunz (approved):

PR #58, opened 2021-01-19
Add support for open-sourced JuiceFS

### LINKED ISSUE #8: JuiceFS CSI plugin for Kubernetes
**What would you like to be added**:
[CSI](https://github.com/container-storage-interface/spec/blob/master/spec.md) plugin for JuiceFS

**Why is this needed**:
As more and more applications are running on Kubernetes, CSI plugin is the de facto form to use third party storage like JuiceFS.

chnliyong:
1. Add support for community JuiceFS (fix https://github.com/juicedata/juicefs/issues/8)
2. Remove cluster-driver-registrar sidecar as it has already [deprecated](https://kubernetes-csi.github.io/docs/cluster-driver-registrar.html)
3. Upgrade `csi-provisioner` to version `v1.6.0` as Kubernetes 1.20 [Stop propagating SelfLink](https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG/CHANGELOG-1.20.md)

### DIFF
--- a/examples/basic/README.md
+++ b/examples/basic/README.md
@@ -4,19 +4,19 @@ This example shows how a basic example to use JuiceFS in Kubernetes pod.
 
 ## Prerequisite
 
-Create JuiceFS filesystem in [JuiceFS web console](https://juicefs.com/console)
+Create JuiceFS filesystem using [juicefs format](https://github.com/juicedata/juicefs/#format-a-volume). Take Amazon S3 as an example:
+
+```
+./juicefs format --storage=s3 --bucket=https://<BUCKET>.s3.<REGION>.amazonaws.com redis://user:password@<HOST>:6379/1 <NAME>
+```
+
+
 
 Then create secret for access in Kubernetes.
 
 ```sh
->> kubectl -n default create secret generic juicefs-secret --from-literal=name=$JFS_NAME --from-literal=token=$JFS_TOKEN --from-literal=accesskey=$JFS_ACCESSKEY --from-literal=secretkey=$JFS_SECRETKEY
+>> kubectl -n default create secret generic juicefs-secret --from-literal=metaurl=redis://user:password@<HOST>:6379/1
 ```
-
-JuiceFS token can be found in `https://juicefs.com/console/vol/{name}/setting`
-
-**Note**: If you are using Tencent Cloud COS, a bucket name with `AppID` as suffix is required.  
-You can find this bucket name in `https://juicefs.com/console/vol/{name}/setting`, `Basic Information` section.  
-Add bucket name like `--from-literal=bucket=$JFS_BUCKET_NAME` when creating kubernetes secret.
 
 ## Apply the Example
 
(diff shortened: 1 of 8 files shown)


examples/basic/README.md
```diff
@@ -4,20 +4,20 @@ This example shows how a basic example to use JuiceFS in Kubernetes pod.
...
+./juicefs format --storage=s3 --bucket=https://<BUCKET>.s3.<REGION>.amazonaws.com redis://user:password@<HOST>:6379/1 <NAME>
+```
+
+
 
 Then create secret for access in Kubernetes.
 
 ```sh
->> kubectl -n default create secret generic juicefs-secret --from-literal=name=$JFS_NAME --from-literal=token=$JFS_TOKEN --from-literal=accesskey=$JFS_ACCESSKEY --from-literal=secretkey=$JFS_SECRETKEY
+>> kubectl -n default create secret generic juicefs-secret --from-literal=metaurl=redis://user:password@<HOST>:6379/1
```
yujunz, line 18: Does it drop support for enterprise edition?

pkg/juicefs/juicefs.go
```diff
@@ -112,14 +116,24 @@ func (j *juicefs) IsNotMountPoint(dir string) (bool, error) {
...
-		return nil, status.Errorf(codes.Internal, "Could not auth juicefs: %v", err)
+	source, ok := secrets["metaurl"]
+	if !ok {
+		j.Upgrade()
+		stdoutStderr, err := j.AuthFs(secrets)
+		klog.V(5).Infof("MountFs: authentication output is '%s'\n", stdoutStderr)
+		if err != nil {
+			return nil, status.Errorf(codes.Internal, "Could not auth juicefs: %v", err)
+		}
+		source = secrets["name"]
```
yujunz, line 127: What is the reason for renaming `name` to `source` while keeping secret key not unchanged?

examples/basic/README.md
chnliyong: The enterprise edition is also supported. The docs need to be updated later(WIP), and only for community JuiceFS here, enterprise docs are placed on juicefs.com.

pkg/juicefs/juicefs.go
chnliyong: The `source` name from mount command like `mount -t fstype source target`. The community edition use _redis url_ as the `source` parameter, it seems not like a `name`.

yujunz: Do we need to change the key name in secret from `name` to `source` for consistency? Or making an alias or something to make it self explainable.

chnliyong: Good idea. Maybe we also need move some secret out from secrets to storage class parameters like `bucket`, `storage` etc.

davies: It's better to use `source` here, which could be redis-url or name of volume in juicefs.com.

Issue #60, opened 2021-01-21
Install juicefs-csi-driver throw kind "CSIDriver" in version "storage.k8s.io/v1" on k8s 1.15


wph95:
as the document said, CSI-driver support k8s version v1.13+
btw in k8s 1.14/1.15 CSI feature is Beta
so need an install file which 

```
apiVersion: storage.k8s.io/v1beta1
# apiVersion: storage.k8s.io/v1 

kind: CSIDriver
```

chnliyong:
>  so need an install file which

Yeah, you are right. We'll do more Kubernetes version test. Thanks!

davies:
So this should be fixed, thanks!

PR #61, opened 2021-01-21
change csidriver apiversion for old kubernetes version

chnliyong:
Related issue: #60 

As Kubernetes 1.13, the CSIDriver is not [enabled by default](https://kubernetes-csi.github.io/docs/csi-driver-object.html#enabling-csidriver-on-kubernetes), and the `apiVersion` is `storage.k8s.io/v1alpha1`, since version 1.14, the `apiVersion` is `storage.k8s.io/v1beta1` and [is not deprecated](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.20/#csidriver-v1beta1-storage-k8s-io) on Kubernetes 1.20(currently the newest version), we change the compatibility from 1.14+ for simplicity.

If we must use Kubernetes 1.13, first [enable CSIDriver](https://kubernetes-csi.github.io/docs/csi-driver-object.html#enabling-csidriver-on-kubernetes), then change the `apiVersion` of `CSIDriver` from `storage.k8s.io/v1beta1` to `storage.k8s.io/v1alpha1` in `deploy/k8s.yaml`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -62,7 +62,7 @@ The following sections are Kubernetes specific. If you are a Kubernetes user, us
 
 ### Kubernetes Version Compatibility
 
-JuiceFS CSI Driver is compatible with Kubernetes **v1.13+**
+JuiceFS CSI Driver is compatible with Kubernetes **v1.14+**
 
 Container Images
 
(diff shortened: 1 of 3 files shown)


davies:
1.14+ should be enough.

PR #69, opened 2021-02-01
Update chart templates

ghost:
Hi, I was trying to deploy this csi driver to my cluster, and found the existing chart is not able to connect in cluster redis instance with `hostNetwork: true` while dnsPolicy kept as default.

In this pull request, I was trying to fix csi-driver deployment and storage class creation:

- Fix csi driver daemonset dns policy for cluster service
- Create secret template for storage class backend


chnliyong:
Good job!

PR #87, opened 2021-04-20
Add troubleshooting document

xiaogaozi:
- Add troubleshooting document
- Change default value of `reclaimPolicy` from `Delete` to `Retain`


README.md
```diff
@@ -48,7 +48,47 @@ helm repo update
 helm upgrade juicefs-csi-driver juicefs-csi-driver/juicefs-csi-driver --install -f ./values.yaml
 ```
 
-3. After above steps, `juicefs-sc` storage class is created.
+3. Check the deployment
+
+- Check pods are running: the deployment will launch a `StatefulSet` named `juicefs-csi-controller` with replica `1` and a `DaemonSet` named `juicefs-csi-node`, so run `kubectl -n kube-system get pods | grep juicefs-csi` should see `n+1` (where `n` is the number of worker nodes of the Kubernetes cluster) pods is running. For example:
```
yujunz: We may select pods by label which will be more precise 

See https://github.com/juicedata/juicefs-csi-driver/blob/21939d0e3fd9adb865b2e816181df833bd5a64c5/deploy/kubernetes/base/kustomization.yaml#L8

yujunz (approved):

README.md
xiaogaozi: Good suggestion!

yujunz (approved):

PR #91, opened 2021-05-16
Make NodeUnpublishVolume idempotent

chnliyong:
If the target path for `NodeUnpublishVolume()` isn't a mountpoint, the umount will fail and contiguously retry.
This PR will fix this.

Upgrade mount utils pkg `k8s.io/utils/mount` as it provides some useful API such as `IsCorruptedMnt()`.


pkg/driver/node.go
```diff
@@ -110,7 +111,7 @@ func (d *nodeService) NodePublishVolume(ctx context.Context, req *csi.NodePublis
 	}
 
 	klog.V(5).Infof("NodePublishVolume: creating dir %s", target)
-	if err := d.juicefs.MakeDir(target); err != nil {
+	if err := os.MkdirAll(target, os.FileMode(0755)); err != nil {
```
yujunz, line 114: I vaguely remember the abstract driver layer is used to mock API for testing. Does this change affect the unit tests? If not, please ignore.

pkg/juicefs/juicefs.go
```diff
@@ -70,13 +71,13 @@ func (fs *jfs) CreateVol(volumeID, subPath string) (string, error) {
 	volPath := filepath.Join(fs.MountPath, subPath)
 
 	klog.V(5).Infof("CreateVol: checking %q exists in %v", volPath, fs)
-	exists, err := fs.Provider.ExistsPath(volPath)
+	exists, err := mount.PathExists(volPath)
```
yujunz, line 74: ditto

pkg/driver/node.go
chnliyong: As I upgrade the mount utils package(`k8s.io/utils/mount`), the `fakeJfsProvider` in sanity test also use the `mount.FakeMounter` from `k8s.io/utils/mount`, and the test passed.

davies (approved):

Issue #16, opened 2019-11-02
Support GCP credentials for GS access


yujunz:
Google Storage requires credentials files referred by environment variable. Currently JuiceFS CSI driver supports command line arguments only for authentication only.

Need a solution for file bases auth.

davies:
fixed

PR #114, opened 2021-07-20
support add config & env in mount pod

zwwhdls:
1. support add config & env in mount pod
2. add example of config and env

support https://github.com/juicedata/juicefs-csi-driver/issues/16

### DIFF
--- a/pkg/juicefs/pod.go
+++ b/pkg/juicefs/pod.go
@@ -109,32 +109,28 @@ func NewMountPod(podName, cmd, mountPath string, resourceRequirements corev1.Res
 	if JFSMountPriorityName != "" {
 		pod.Spec.PriorityClassName = JFSMountPriorityName
 	}
-	if configs != nil {
-		i := 1
-		for k, v := range configs {
-			pod.Spec.Containers[0].VolumeMounts = append(pod.Spec.Containers[0].VolumeMounts,
-				corev1.VolumeMount{
-					Name:      fmt.Sprintf("config-%v", i),
-					MountPath: v,
-				})
-			pod.Spec.Volumes = append(pod.Spec.Volumes, corev1.Volume{
-				Name: fmt.Sprintf("config-%v", i),
-				VolumeSource: corev1.VolumeSource{
-					Secret: &corev1.SecretVolumeSource{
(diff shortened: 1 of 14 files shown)


pkg/juicefs/secret.go
```diff
@@ -0,0 +1,82 @@
...
+package juicefs
+
+import (
+	"encoding/json"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+	"k8s.io/apimachinery/pkg/util/yaml"
+)
+
+type JfsSecret struct {
```
chnliyong: The fields are not all secret, may `JfsConfig` or `JfsSetting` more suitable？

pkg/juicefs/secret.go
```diff
@@ -0,0 +1,82 @@
...
+	Configs map[string]string `json:"configs"`
+	Envs    map[string]string `json:"envs"`
+
+	MountPodCpuLimit   string `json:"mount_pod_cpu_limit"`
+	MountPodMemLimit   string `json:"mount_pod_mem_limit"`
+	MountPodCpuRequest string `json:"mount_pod_cpu_request"`
+	MountPodMemRequest string `json:"mount_pod_mem_request"`
+}
+
+func ParseSecret(secrets, volCtx map[string]string) (*JfsSecret, error) {
```
chnliyong: `volCtx` isn't secret, `ParseJfsConfig`

pkg/juicefs/secret.go
zwwhdls: ok

pkg/juicefs/pod.go
```diff
@@ -102,6 +109,33 @@ func NewMountPod(podName, cmd, mountPath string, resourceRequirements corev1.Res
...
+				VolumeSource: corev1.VolumeSource{
+					Secret: &corev1.SecretVolumeSource{
+						SecretName: k,
+					},
+				},
+			})
+			i++
+		}
+	}
+	if env != nil {
```
davies: nit: it's safe to iterator on nil map

zwwhdls: got

PR #116, opened 2021-07-22
fix csi controller

zwwhdls:
fix controller DeleteVolume & update controller yaml


charts/latest/juicefs-csi-driver/values.yaml
```diff
@@ -9,7 +9,7 @@
 
 image:
   repository: juicedata/juicefs-csi-driver
-  tag: "latest"
+  tag: "nightly"
```
chnliyong: `nightly` only for our test, keep `latest` when publish

chnliyong (approved):

Issue #164, opened 2021-09-10
Improvement for helm chart's output message.


matrixji:
I'm installing csi-driver chart with below commands:
```bash
helm upgrade juicefs-csi-driver  juicefs-csi-driver/juicefs-csi-driver --install -f  juicefs_values.yaml
```

and the content of juicefs_values.yaml below:
```yaml
- name: juicefs
  enabled: true
  reclaimPolicy: Retain
  backend:
    name: juicefs
    metaurl: redis://:redis@redis-master.redis.svc
    storage: s3
    accessKey: minio
    secretKey: minio123
    bucket: http://minio-service.default.svc:9000/milvus-test
```

I'm using `juicefs` as the `name` parameter. and I've got the output for test if juicefs is working correctly like below:
```

...

apiVersion: v1
kind: PersistentVolumeClaim
metadata:
  name: juicefs-pvc
  namespace: default
spec:
  accessModes:
  - ReadWriteMany
  resources:
    requests:
      storage: 10Pi
  storageClassName: juicefs-sc

...

```

the example using `juicefs-sc` as the storageClassName, but indeed, we create it with the name with `juicefs`.

zwwhdls:
fixed by https://github.com/juicedata/juicefs-csi-driver/pull/165

PR #165, opened 2021-09-10
Improvment: using storageClassName in example code from values.

matrixji:
- using user input `name` parameter in `storageClasses` for example code
  in output message.

ref: #164.

Signed-off-by: Ji Bin <matrixji@live.com>


zwwhdls:
LGTM, thanks!

PR #181, opened 2021-10-20
Allow mount options to be configurable

### LINKED ISSUE #178: JFS Default Cache should be configurable in Helm Chart
It would be nice if this config was configurable in the values.yaml file for the Helm chart.

Useful to have a safe default e.g. /tmp if the user doesn't specify it in the mountOptions.

```
jfs-default-cache:                                                                                                                                                                                                                      
  Type:          HostPath (bare host directory volume)                                                                                                                                                                                  
  Path:          /var/jfsCache
  HostPathType:  DirectoryOrCreate
```

threeseed:
Fixes #178.

### DIFF
--- a/charts/juicefs-csi-driver/values.yaml
+++ b/charts/juicefs-csi-driver/values.yaml
@@ -139,7 +139,7 @@ storageClasses:
 
 
   mountOptions:
-  # -- Mount Options. Read [this document](https://github.com/juicedata/juicefs/blob/main/docs/en/cache_management.md) to learn how to set mount options.
+  # -- Mount Options. Read [this document](https://github.com/juicedata/juicefs/blob/main/docs/en/command_reference.md#juicefs-mount) to learn how to set different mount options.
 
   mountPod:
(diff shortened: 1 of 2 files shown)


charts/juicefs-csi-driver/values.yaml
```diff
@@ -137,6 +137,10 @@ storageClasses:
     # -- Bucket URL. Read [this document](https://github.com/juicedata/juicefs/blob/main/docs/en/how_to_setup_object_storage.md) to learn how to setup different object storage.
     bucket: ""
 
+
+  mountOptions:
+  # -- Mount Options. Read [this document](https://github.com/juicedata/juicefs/blob/main/docs/en/cache_management.md) to learn how to set mount options.
```
davies: Please use this link: https://github.com/juicedata/juicefs/blob/main/docs/en/command_reference.md#juicefs-mount

davies (changes requested):

davies (approved):

PR #185, opened 2021-10-26
add volume recovery

ExponTec:
Contributed by Expontech

### DIFF
--- a/pkg/controller/pod_driver.go
+++ b/pkg/controller/pod_driver.go
@@ -28,8 +28,10 @@ import (
 	"k8s.io/klog"
 	k8sMount "k8s.io/utils/mount"
 	"os"
+	"os/exec"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+	"strings"
 	"time"
 )
 
@@ -151,6 +153,10 @@ func (p *PodDriver) podErrorHandler(ctx context.Context, pod *corev1.Pod) (recon
 }
 
 func (p *PodDriver) podDeletedHandler(ctx context.Context, pod *corev1.Pod) (reconcile.Result, error) {
+	if pod == nil {
+		klog.Errorf("get nil pod")
+		return reconcile.Result{}, nil
+	}
 	klog.V(5).Infof("Get pod %s in namespace %s is to be deleted.", pod.Name, pod.Namespace)
 	if !util.ContainsString(pod.GetFinalizers(), config.Finalizer) {
 		// do nothing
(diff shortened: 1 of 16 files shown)


pkg/controller/pod_driver.go
```diff
@@ -86,19 +100,122 @@ func (p *PodDriver) podDeletedHandler(ctx context.Context, pod *corev1.Pod) (rec
...
+	if err := p.Mounter.Unmount(sourcePath); err != nil {
+		klog.Errorf("umount %s err:%v\n", sourcePath, err)
+		//return reconcile.Result{}, nil
+	}
+	// create
+	klog.V(5).Infof("pod targetPath not empty, need create thd pod:%s", pod.Name)
+	// todo check pod not exists
+	controllerutil.AddFinalizer(pod, config.Finalizer)
+	pod.ResourceVersion = ""
+	err := p.Client.Create(ctx, pod)
```
zwwhdls: Can we new a pod and use old pod spec? And also check old pod is deleted before create

pkg/juicefs/mount/pod.go
```diff
@@ -104,15 +107,17 @@ func NewMountPod(podName, cmd, mountPath string, resourceRequirements corev1.Res
...
 	var pod = &corev1.Pod{
 		ObjectMeta: metav1.ObjectMeta{
 			Name:      podName,
 			Namespace: config.Namespace,
 			Labels: map[string]string{
 				config.PodTypeKey: config.PodTypeValue,
 			},
-			Annotations: make(map[string]string),
+			Annotations: map[string]string{
+				config.VolumeIdKey: volId,
```
zwwhdls: Adding new annotation means recovery didn't work for existing mount pod.

zwwhdls:
If the pod fails to deploy due to insufficient resources, csi node will delete resources in pod and deploy it again. This will conflict with errorHandler in manager.

zwwhdls:
@ExponTec hi, would you fix Travis CI please?

pkg/util/pod.go
```diff
@@ -21,6 +21,13 @@ import (
 	"strings"
 )
 
+func PodReadyStatus(pod *corev1.Pod) bool {
+	if pod.Status.Phase == corev1.PodRunning && !podNotRunning(pod.Status.ContainerStatuses) {
```
zwwhdls: If containerReady is not true, this function will return true, it dose not meet expectations. You can use IsPodReady function instead

zwwhdls:
Hi @ExponTec , 

Would you add some unit test?

pkg/controller/pod_driver.go
```diff
@@ -86,19 +163,167 @@ func (p *PodDriver) podDeletedHandler(ctx context.Context, pod *corev1.Pod) (rec
...
+		return reconcile.Result{}, nil
+	}
+	sourcePath, _, err := util.ParseMntPath(cmd[2])
+	if err != nil {
+		klog.Error(err)
+		return reconcile.Result{}, nil
+	}
+
+	klog.Infof("start umount :%s", sourcePath)
+	if err := p.Mounter.Unmount(sourcePath); err != nil {
```
zwwhdls: If mountPoint in node is umounted manually,  mount pod will be error and then be deleted by manager, but umount function here will be error.

PR #187, opened 2021-10-30
Fix  the environment variables key error  of use juicefs-csi-driver in GKE

zhijian-pro:
Change the environment variables key  `GKEGOOGLE_CLOUD_PROJECT` to `GOOGLE_APPLICATION_CREDENTIALS`


zwwhdls:
Please also update GKE env in example.

PR #234, opened 2021-12-23
Configure 'trash-days' in format

lixuchun314:
增加--trash-days参数作用在juicefs format的时候


pkg/juicefs/juicefs.go
```diff
@@ -429,6 +429,7 @@ func (j *juicefs) ceFormat(secrets map[string]string, noUpdate bool, extraEnvs m
 		"access-key",
 		"block-size",
 		"compress",
+		"trash-days",
```
zwwhdls, line 432: `trash-days` is optional, please add it in `isOptional` map also.

PR #240, opened 2021-12-30
fix publish when mount pod deleted delay

### LINKED ISSUE #239: [BUG] fake Mount success when mount pod deleting has a long delay
**What happened**:

1. when when all the user pod of mount pod being deleted, the mount pod will be deleted followed
2. mount pod is deleted status, but it still exists on node, it may take some time to destroy. Here has a time gap.
3. In the destroy time gap, mountSource path still exist
4. new user pod created, NodePublishVolume invoked
5. CSI Node first checks the mountSource, and will consider mount pod exists normal incorrectly.
6. CSI Node add refs to mount pod annotation, and return success to kubelet
7. after mount pod really destroyed, the mount of new user pod will be corrupted and could not be recover

**Environment**:
- JuiceFS CSI Driver version (which image tag did your CSI Driver use): v0.11.1
- Others:

zwwhdls:
fix https://github.com/juicedata/juicefs-csi-driver/issues/239

Check if mount pod deleted when adding ref in annotation. If deleting, return err.

### DIFF
--- a/pkg/juicefs/mount/pod_mount.go
+++ b/pkg/juicefs/mount/pod_mount.go
@@ -235,6 +235,9 @@ func (p *PodMount) AddRefOfMount(target string, podName string) error {
 		if err != nil {
 			return err
 		}
+		if exist.DeletionTimestamp != nil {
+			return status.Errorf(codes.Internal, "addRefOfMount: Mount pod [%s] has been deletes.", podName)
+		}
 		annotation := exist.Annotations
 		if _, ok := annotation[key]; ok {
 			klog.V(5).Infof("addRefOfMount: Target ref [%s] in pod [%s] already exists.", target, podName)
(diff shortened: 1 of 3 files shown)


pkg/juicefs/mount/pod_mount.go
```diff
@@ -235,6 +235,9 @@ func (p *PodMount) AddRefOfMount(target string, podName string) error {
 		if err != nil {
 			return err
 		}
+		if exist.DeletionTimestamp != nil {
+			return status.Errorf(codes.Internal, "addRefOfMount: Mount pod [%s] has been deletes.", podName)
```
yujunz:
```suggestion
			return status.Errorf(codes.Internal, "addRefOfMount: Mount pod [%s] has been deleted.", podName)
```

yujunz (changes requested):
Typo?

pkg/juicefs/mount/pod_mount.go
zwwhdls: yes, typo...

davies (approved):

PR #247, opened 2022-01-11
Fix concurrency issue

zwwhdls:
add lock of each pod. use podLock in csi and manager.


cmd/main.go
```diff
@@ -41,6 +41,11 @@ var (
 func init() {
 	klog.InitFlags(nil)
 	flag.Parse()
+	podLocks := make([]*sync.Mutex, 1024)
```
davies: config.PodLocks = make([]sync.Mutex, 1024)

davies: Then we don't need the following lines.

pkg/config/config.go
```diff
@@ -64,3 +64,11 @@ const (
 	mountPodAnnotationKey  = "juicefs/mount-annotations"
 	mountPodServiceAccount = "juicefs/mount-service-account"
 )
+
+var PodLocks []*sync.Mutex
+
+func GetPodLock(podName string) *sync.Mutex {
+	key := md5.Sum([]byte(podName))
```
davies: key is not a number, you may want to use other Hash function

pkg/juicefs/k8sclient/client.go
```diff
@@ -32,6 +39,10 @@ type K8sClient struct {
 
 func NewClient() (*K8sClient, error) {
 	config, err := rest.InClusterConfig()
+	if config == nil {
+		return nil, status.Error(codes.NotFound, "Can't get kube InClusterConfig")
```
davies, line 46: this will hide the above error

pkg/config/config.go
```diff
@@ -64,3 +63,12 @@ const (
 	mountPodAnnotationKey  = "juicefs/mount-annotations"
 	mountPodServiceAccount = "juicefs/mount-service-account"
 )
+
+var PodLocks []sync.Mutex
```
fishincat: Why init PodLocks in test file so many times instead of here?

pkg/controller/pod_driver.go
```diff
@@ -315,21 +374,25 @@ func (p *PodDriver) podReadyHandler(ctx context.Context, pod *corev1.Pod) (recon
...
 
-	// recovery for each target
-	for k, target := range pod.Annotations {
-		if k == util.GetReferenceKey(target) {
-			mi := p.mit.resolveTarget(target)
-			if mi == nil {
-				klog.Errorf("pod %s target %s resolve fail", pod.Name, target)
-				continue
-			}
+	_ = doWithinTime(ctx, func() error {
```
fishincat: This function can not be executed in concurrence!
Use doWithinTime will case some unexpeced things happen.

pkg/juicefs/mount/pod_mount.go
```diff
@@ -173,64 +183,60 @@ func (p *PodMount) waitUntilMount(jfsSetting *jfsConfig.JfsSetting, volumeId, ta
...
 
+func (p *PodMount) waitUtilPodReady(volumeId string) error {
+	podName := GeneratePodNameByVolumeId(volumeId)
 	// Wait until the mount pod is ready
-	for i := 0; i < 30; i++ {
+	for i := 0; i < 60; i++ {
 		pod, err := p.K8sClient.GetPod(podName, jfsConfig.Namespace)
 		if err != nil {
-			return status.Errorf(codes.Internal, "waitUtilMount: Get pod %v failed: %v", volumeId, err)
+			return status.Errorf(codes.Internal, "waitUtilPodReady: Get pod %v failed: %v", volumeId, err)
```
davies, line 223: The pod could be deleted and recovered, so we should retry when error is returned

pkg/juicefs/mount/pod_mount.go
```diff
@@ -173,64 +183,60 @@ func (p *PodMount) waitUntilMount(jfsSetting *jfsConfig.JfsSetting, volumeId, ta
...
+				// pod not exist, create
+				klog.V(5).Infof("createOrAddRef: Need to create pod %s.", podName)
+				newPod := NewMountPod(podName, cmd, mountPath, podResource, config, env, labels, anno, serviceAccount)
+				if newPod.Annotations == nil {
+					newPod.Annotations = make(map[string]string)
+				}
+				newPod.Annotations[key] = target
+				_, err = p.K8sClient.CreatePod(newPod)
+				if err != nil {
+					klog.Errorf("createOrAddRef: Create pod %s err and retry: %v", podName, err)
```
davies, line 203: continue?

pkg/juicefs/mount/pod_mount.go
```diff
@@ -173,64 +183,60 @@ func (p *PodMount) waitUntilMount(jfsSetting *jfsConfig.JfsSetting, volumeId, ta
...
+				}
+				return nil
 			}
-			klog.Errorf("get pod err:%v", err)
+			// unexpect error
+			klog.Errorf("createOrAddRef: Get pod %s err and retry: %v", podName, err)
```
davies, line 208: continue?

pkg/juicefs/mount/pod_mount.go
```diff
@@ -173,64 +183,60 @@ func (p *PodMount) waitUntilMount(jfsSetting *jfsConfig.JfsSetting, volumeId, ta
...
-		newPod := NewMountPod(podName, cmd, mountPath, podResource, config, env, labels, anno, serviceAccount)
-		if newPod.Annotations == nil {
-			newPod.Annotations = make(map[string]string)
-		}
-		newPod.Annotations[key] = target
-		if _, e := p.K8sClient.CreatePod(newPod); e != nil {
-			return e
-		}
+		// pod exist, add refs
+		return p.AddRefOfMount(target, podName)
```
davies, line 212: retry if failed to add ref

pkg/juicefs/mount/pod_mount.go
```diff
@@ -43,7 +43,10 @@ func NewPodMount(client *k8sclient.K8sClient, mounter k8sMount.SafeFormatAndMoun
 }
 
 func (p *PodMount) JMount(jfsSetting *jfsConfig.JfsSetting, volumeId, mountPath string, target string, options []string) error {
-	return p.waitUntilMount(jfsSetting, volumeId, target, mountPath, p.getCommand(jfsSetting, mountPath, options))
+	if err := p.createOrAddRef(jfsSetting, volumeId, target, mountPath, p.getCommand(jfsSetting, mountPath, options)); err != nil {
+		return err
```
davies, line 47: Should we call JUmount to cleanup?

pkg/juicefs/mount/pod_mount.go
zwwhdls: Call apiServer error is unexpected error, we should return it and wait for kubelet retry

pkg/juicefs/mount/pod_mount.go
zwwhdls: ditto

pkg/juicefs/mount/pod_mount.go
zwwhdls: ditto

pkg/juicefs/mount/pod_mount.go
zwwhdls: ditto

PR #260, opened 2022-01-22
Fix pod leak

zwwhdls:
Check annotation refs in mount pod in driver. If target pod doesn't exist, delete ref and if no refs in mount pod, delete mount pod.

### DIFF
--- a/pkg/controller/pod_driver.go
+++ b/pkg/controller/pod_driver.go
@@ -84,6 +84,11 @@ func (p *PodDriver) Run(ctx context.Context, current *corev1.Pod) error {
 	if err := p.preCheck(pod); err != nil {
 		return err
 	}
+	// get latest pod resourceVersion from apiserver after update refs
+	pod, err = p.Client.GetPod(current.Name, current.Namespace)
+	if err != nil {
+		return nil
+	}
 	return p.handlers[p.getPodStatus(pod)](ctx, pod)
 }
 
--- a/pkg/controller/pod_driver.go
+++ b/pkg/controller/pod_driver.go
@@ -23,6 +23,7 @@ import (
 	"google.golang.org/grpc/status"
 	"os"
 	"os/exec"
+	"reflect"
 	"strings"
 	"time"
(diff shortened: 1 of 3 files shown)


pkg/controller/pod_driver.go
```diff
@@ -104,6 +121,44 @@ func (p *PodDriver) getPodStatus(pod *corev1.Pod) podStatus {
...
+func (p *PodDriver) checkAnnotations(pod *corev1.Pod) error {
+	// check refs in mount pod, the corresponding pod exists or not
+	lock := config.GetPodLock(pod.Name)
+	lock.Lock()
+	defer lock.Unlock()
+
+	annotation := make(map[string]string)
+	var existTargets int
+	for k, target := range pod.Annotations {
+		if strings.HasPrefix(k, "juicefs-") {
```
fishincat, line 133: should check it valid like:
```
for k, target := range pod.Annotations {
		if k == util.GetReferenceKey(target) {
```

pkg/controller/pod_driver.go
```diff
@@ -104,6 +121,44 @@ func (p *PodDriver) getPodStatus(pod *corev1.Pod) podStatus {
...
+			deleted, exists := p.mit.deletedPods[getPodUid(target)]
```
fishincat, line 134: exists is false should consider it not deleted.

pkg/controller/pod_driver.go
```diff
@@ -104,6 +121,44 @@ func (p *PodDriver) getPodStatus(pod *corev1.Pod) podStatus {
...
+				continue
+			}
+			existTargets++
+		}
+		annotation[k] = target
+	}
+
+	if len(pod.Annotations) != len(annotation) {
+		pod.Annotations = annotation
+		if err := p.Client.UpdatePod(pod); err != nil {
```
fishincat, line 146: UpdatePod may fail because resourceVersion of pod in kubelet may be behind to the apiServive.

PR #262, opened 2022-01-24
put format/auth in initcontainer

zwwhdls:


pkg/juicefs/mount/resources/secret.go
```diff
@@ -0,0 +1,84 @@
...
+	if setting.EncryptRsaKey != "" {
+		data["encrypt_rsa_key"] = setting.EncryptRsaKey
+	}
+	if setting.InitConfig != "" {
+		data["init_config"] = setting.InitConfig
+	}
+	for k, v := range setting.Envs {
+		data[k] = v
+	}
+	klog.V(6).Infof("secret data: %v", data)
```
davies: this is dangerous, we should not log any secrets.

pkg/juicefs/mount/resources/common.go
```diff
@@ -40,12 +56,25 @@ func generateJuicePod(jfsSetting *config.JfsSetting) *corev1.Pod {
...
+	if jfsSetting.FormatCmd != "" {
+		initContainer := getInitContainer(jfsSetting)
+		initContainer.VolumeMounts = append(initContainer.VolumeMounts, volumeMounts...)
+		pod.Spec.InitContainers = []corev1.Container{initContainer}
+	}
 	return pod
 }
 
-func getVolumes() []corev1.Volume {
+func getVolumes(setting config.JfsSetting) []corev1.Volume {
```
davies: it's better to pass by reference for large object

pkg/juicefs/juicefs.go
```diff
@@ -259,81 +287,61 @@ func (j *juicefs) AuthFs(secrets map[string]string, extraEnvs map[string]string)
...
-	envs := syscall.Environ()
-	for key, val := range extraEnvs {
-		envs = append(envs, fmt.Sprintf("%s=%s", key, val))
 	}
-	klog.V(5).Infof("AuthFs: envs %q", extraEnvs)
-	authCmd.SetEnv(envs)
-	return authCmd.CombinedOutput()
+
+	cmd := strings.Join(args, " ")
+	klog.V(5).Infof("AuthFs: cmd %s", cmd)
```
davies: There are secrets in cmd

pkg/juicefs/k8sclient/client.go
```diff
@@ -77,6 +81,44 @@ func (k *K8sClient) GetPod(podName, namespace string) (*corev1.Pod, error) {
...
+	if err != nil {
+		klog.V(6).Infof("Can't list pod in namespace %s by labelSelector %v: %v", namespace, labelSelector, err)
+		return nil, err
+	}
+	return podList.Items, nil
+}
+
+func (k *K8sClient) GetPodLog(podName, namespace, containerName string) (string, error) {
+	klog.V(6).Infof("Get pod %s log", podName)
+	tailLines := int64(5)
```
davies: 20 lines?

davies (changes requested):
1. make sure no secret in logging
2. Keep the previous behavior (process)
3. Add a feature flag for format/auth in container (off by default, on in tests)

pkg/juicefs/juicefs.go
zwwhdls: Secret in this cmd is replaced by string `$(xxx)`, this log will not display secret

chnliyong (approved):

Issue #270, opened 2022-02-05
Support for Hashicorp Nomad


OmegaRogue:
**What feature you'd like to add:**
Added support for running on [HashiCorp Nomad](https://www.nomadproject.io/)
**Why is this feature needed:**
Nomad supports the CSI specification and allows for a wider variety of workloads than kubernetes.
I have tried to run this driver on nomad, but the direct kubernetes dependencies are causing errors.

davies:
The docs from Monad [1] says the K8S CSI driver work out of box with Monad, what the error you saw?

This CSI driver relies on k8S POD, does Monad has similar feature to manage containers?

[1] https://www.nomadproject.io/docs/internals/plugins/csi

OmegaRogue:
On the controller allocation:
```
F0206 20:08:51.194502       8 main.go:50] Pod name & namespace can't be null.
```

On the node allocations:
```
Failed to start container c7d3a3b472de596fcc2c7986588f73bb6f7154c99a30f65d33c1f0e8f1ae31c2: API error (500): path /opt/nomad-client/data/client/csi/node/juicefs is mounted on / but it is not a shared mount
```
Nomad supports the Container Storage Interface, but not the kubernetes API, and this driver is explicitly dependant on both

OmegaRogue:
Nomad manages a variety of different workloads, I'm running the allocations in Docker containers. This blog post outlines the similarities between kubernetes and nomad https://www.hashicorp.com/blog/a-kubernetes-user-s-guide-to-hashicorp-nomad

OmegaRogue:
Any update on this?

davies:
@OmegaRogue Can you try the versions before 0.10? Which may not use K8S API.

OmegaRogue:
``
F0221 14:34:46.483746       1 main.go:47] nodeID must be provided
goroutine 1 [running]:
k8s.io/klog.stacks(0xc0001c5100, 0xc00017a500, 0x42, 0x98)
	/go/pkg/mod/k8s.io/klog@v0.1.0/klog.go:828 +0xb8
k8s.io/klog.(*loggingT).output(0xf92e20, 0xc000000003, 0xc0002644d0, 0xf2aecc, 0x7, 0x2f, 0x0)
	/go/pkg/mod/k8s.io/klog@v0.1.0/klog.go:779 +0x351
[...]

PR #293, opened 2022-03-08
Support process mount

zwwhdls:
support https://github.com/juicedata/juicefs-csi-driver/issues/270


cmd/main.go
```diff
@@ -35,11 +35,21 @@ var (
 	enableManager      = flag.Bool("enable-manager", false, "Enable manager or not.")
 	reconcilerInterval = flag.Int("reconciler-interval", 5, "interval (default 5s) for reconciler")
 	formatInPod        = flag.Bool("format-in-pod", false, "Put format/auth in pod")
+	inKube             = flag.Bool("in-kube", true, "CSI Driver run in kubernetes or not. default true.")
```
davies: it's better to use `false` as default value.

chnliyong (approved):

PR #318, opened 2022-04-21
remove csi registration sock before stop

### LINKED ISSUE #311: Can we remove the node's "csi.volume.kubernetes.io/nodeid" annotation when uninstall JuiceFS CSI?
**What happened**:

I execute `heml uninstall juicefs-csi-driver -nkube-system` and the node's "csi.volume.kubernetes.io/nodeid" annotation is still preserved

**Environment**:
- Others:

zwwhdls:
fix https://github.com/juicedata/juicefs-csi-driver/issues/311

### DIFF
--- a/deploy/k8s.yaml
+++ b/deploy/k8s.yaml
@@ -400,6 +400,13 @@ spec:
         - name: JUICEFS_CONFIG_PATH
           value: /var/lib/juicefs/config
         image: juicedata/juicefs-csi-driver:v0.13.2
+        lifecycle:
+          preStop:
+            exec:
+              command:
+              - /bin/sh
+              - -c
+              - rm -rf /registration/csi.juicefs.com-reg.sock
         livenessProbe:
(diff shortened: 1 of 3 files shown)


deploy/k8s.yaml
```diff
@@ -400,6 +400,13 @@ spec:
         - name: JUICEFS_CONFIG_PATH
           value: /var/lib/juicefs/config
         image: juicedata/juicefs-csi-driver:v0.13.2
+        lifecycle:
+          preStop:
+            exec:
+              command:
+              - /bin/sh
+              - -c
+              - rm -rf /registration/csi.juicefs.com-reg.sock /csi/csi.sock
```
yujunz, line 409: Do we really need `-rf`?

zwwhdls: You are right, `-rf` is not needed, clear it.

PR #322, opened 2022-04-22
add updatedb volume mount

### LINKED ISSUE #320: Mount `/etc/updatedb.conf` file in K8s node into mount pod

If the `/etc/updatedb.conf` file exists on the K8s node, mount it in the Mount Pod.

[`updatedb`](https://www.man7.org/linux/man-pages/man1/updatedb.1.html) is a tool to generate an index database for local files, which will be used by the `locate` command. In some Linux distributions (such as Ubuntu), the `updatedb` command is executed regularly every day (you can check the `/etc/cron.daily` directory to confirm), if the current node has JuiceFS mounted, it will also scan the entire JuiceFS file system, which creates an unnecessary load on the metadata engine of the JuiceFS file system. The JuiceFS file system can be ignored by the `updatedb` command by modifying the [`/etc/updatedb.conf`](https://linux.die.net/man/5/updatedb.conf) file.

The JuiceFS client will [detect](https://github.com/juicedata/juicefs/blob/main/cmd/mount_unix.go#L134) by default whether the `/etc/updatedb.conf` file exists on the current node, and if so, then add the appropriate configuration to make the `updatedb` command ignore the JuiceFS file system. But in the K8s environment, the client running on the Mount Pod cannot view the `/etc/updatedb.conf` file of the current node, so it cannot be modified.

zwwhdls:
fix https://github.com/juicedata/juicefs-csi-driver/issues/320

### DIFF
--- a/pkg/juicefs/mount/builder/common.go
+++ b/pkg/juicefs/mount/builder/common.go
@@ -71,6 +71,7 @@ func (r *Builder) generateJuicePod() *corev1.Pod {
 
 func (r *Builder) getVolumes() []corev1.Volume {
 	dir := corev1.HostPathDirectoryOrCreate
+	file := corev1.HostPathFileOrCreate
 	secretName := r.jfsSetting.SecretName
 	volumes := []corev1.Volume{{
 		Name: "jfs-dir",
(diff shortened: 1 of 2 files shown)


pkg/juicefs/mount/builder/common.go
```diff
@@ -86,6 +87,13 @@ func (r *Builder) getVolumes() []corev1.Volume {
 				Path: config.JFSConfigPath,
 				Type: &dir,
 			},
+		}}, {
+		Name: "updatedb",
+		VolumeSource: corev1.VolumeSource{
+			HostPath: &corev1.HostPathVolumeSource{
+				Path: "/etc/updatedb.conf",
```
davies, line 94: If there is no /etc/updatedb.conf in host, the mount pod will fail, right?

zwwhdls: `volumeSource.Type` is set `HostPathFileOrCreate `. If there is no /etc/updatedb.conf in host, it will be created. Mount pod won't fail.

PR #347, opened 2022-06-08
fix mount pod leak when csi node is gone

### LINKED ISSUE #340: [BUG] Mount pod leak when deleting node
**What happened**:
When I delete node pool, many mount pods retained `Terminating` status:

image

That's because CSI node is gone when node pool deleted, finalizer of mount pod can not be remove.

When deleting node pool, mount pods are removed too.

**Environment**:
- Others:

zwwhdls:
fix https://github.com/juicedata/juicefs-csi-driver/issues/340


pkg/controller/mount_controller.go
```diff
@@ -0,0 +1,81 @@
...
+func (m MountController) Reconcile(ctx context.Context, request reconcile.Request) (reconcile.Result, error) {
+	klog.V(6).Infof("Receive pod %s %s", request.Name, request.Namespace)
+	mountPod, err := m.GetPod(request.Name, request.Namespace)
+	if err != nil && !k8serrors.IsNotFound(err) {
+		klog.Errorf("get pod %s error: %v", request.Name, err)
+		return reconcile.Result{}, err
+	}
+	if mountPod == nil {
+		klog.V(6).Infof("pod %s has been deleted.", request.Name)
+		return reconcile.Result{}, err
```
davies: return nil

pkg/controller/mount_controller.go
```diff
@@ -0,0 +1,81 @@
...
+	}
+
+	// check mount pod deleted
+	if mountPod.DeletionTimestamp == nil {
+		klog.V(6).Infof("pod %s is not deleted", mountPod.Name)
+		return reconcile.Result{}, err
```
davies: return nil

pkg/controller/mount_controller.go
```diff
@@ -0,0 +1,81 @@
...
+		// do nothing
+		return reconcile.Result{}, nil
+	}
+
+	// check csi node exist or not
+	nodeName := mountPod.Spec.NodeName
+	labelSelector := metav1.LabelSelector{
+		MatchLabels: map[string]string{config.CSINodeLabelKey: config.CSINodeLabelValue},
+	}
+	csiPods, err := m.ListPod(config.Namespace, labelSelector)
```
davies: Will it be expensive?

cmd/main.go
```diff
@@ -113,7 +117,20 @@ func main() {
...
+		klog.V(5).Infof("Reconciler Started")
+	}
+	if config.MountController {
+		mgr, err := app.NewMountManager()
+		if err != nil {
+			klog.Fatalln(err)
+		}
+		go func() {
+			if err := mgr.Start(ctrl.SetupSignalHandler()); err != nil {
+				klog.Error(err, "fail to run mount controller")
```
davies, line 140: Should this be an warning rather than fatal?

cmd/app/mount_manager.go
```diff
@@ -0,0 +1,83 @@
...
+var (
+	scheme = runtime.NewScheme()
+)
+
+func init() {
+	utilruntime.Must(corev1.AddToScheme(scheme))
+}
+
+func NewMountManager() (ctrl.Manager, error) {
+	var mgr, _ = ctrl.NewManager(ctrl.GetConfigOrDie(), ctrl.Options{
```
davies: don't die

pkg/controller/mount_controller.go
zwwhdls: Add labelSelector and fieldSelector in api, it will be fine.

PR #358, opened 2022-06-23
strip passwd in process mount

### LINKED ISSUE #355: Insecure issue
Hi, I find that the password of meta address doesn't be encrypted, like below:
image
Is it necessary to encrypt it, like juicefs? If yes, let me try to fix it :D

zwwhdls:
fix https://github.com/juicedata/juicefs-csi-driver/issues/355

### DIFF
--- a/pkg/juicefs/mount/process_mount.go
+++ b/pkg/juicefs/mount/process_mount.go
@@ -209,7 +209,7 @@ func (p *ProcessMount) JMount(jfsSetting *jfsConfig.JfsSetting) error {
 		}
 		time.Sleep(time.Millisecond * 500)
 	}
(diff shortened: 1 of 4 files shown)


pkg/juicefs/mount/process_mount.go
```diff
@@ -139,7 +139,7 @@ func (p *ProcessMount) JMount(jfsSetting *jfsConfig.JfsSetting) error {
 		klog.V(5).Infof("eeMount mount success.")
 		return nil
 	}
-	klog.V(5).Infof("ceMount: mount %v at %v", jfsSetting.Source, jfsSetting.MountPath)
+	klog.V(5).Infof("ceMount: mount %v at %v", util.StripPasswd(jfsSetting.Source), jfsSetting.MountPath)
```
davies, line 142: Source is logged in many place, we need update them all.

PR #401, opened 2022-08-25
strip session-token in ceFormat

### LINKED ISSUE #360: [FEATURE] strip session-token in ceFormat

Hexilee:
close #360

### DIFF
--- a/pkg/config/setting.go
+++ b/pkg/config/setting.go
@@ -44,15 +44,14 @@ type JfsSetting struct {
 	CacheDirs     []string   // hostPath using by mount pod
 
 	// put in secret
-	SecretKey           string            `json:"secret-key,omitempty"`
-	SecretKey2          string            `json:"secret-key2,omitempty"`
-	Token               string            `json:"token,omitempty"`
-	Passphrase          string            `json:"passphrase,omitempty"`
-	Envs                map[string]string `json:"envs_map,omitempty"`
-	EncryptRsaKey       string            `json:"encrypt_rsa_key,omitempty"`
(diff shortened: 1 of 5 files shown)


pkg/config/setting.go
```diff
@@ -296,6 +297,53 @@ func ParseSetting(secrets, volCtx map[string]string, options []string, usePod bo
 	return &jfsSetting, nil
 }
 
+func (s JfsSetting) parseFormatOptions() ([][]string, error) {
+	options := strings.Split(s.FormatOptions, ",")
+	parsedOptions := make([][]string, 0)
+	for _, option := range options {
+		pair := strings.Split(strings.TrimSpace(option), "=")
```
zwwhdls, line 304: Some options are not "key=value" form, do not have value.

zwwhdls (changes requested):

pkg/config/setting.go
```diff
@@ -296,6 +297,53 @@ func ParseSetting(secrets, volCtx map[string]string, options []string, usePod bo
 	return &jfsSetting, nil
 }
 
+func (s JfsSetting) parseFormatOptions() ([][]string, error) {
```
zwwhdls: Add unit test please.

pkg/config/setting.go
Hexilee: Single-key options are supported.

pkg/config/setting.go
Hexilee: Added

zwwhdls (approved):

PR #409, opened 2022-08-30
discard local timeout, control the reconciler timeout by context

Hexilee:
There are too many local settings to control the timeout of the whole calling chain, which is unconfigurable and may cause dead goroutine in `kubelet` or controller-runtime.

So, we should only control the timeout of the top context, then we need only to set the global context timeout of controller-runtime (and the `kubelet` sets context timeout by self).

close #403 and #406

PR #414, opened 2022-09-01
pass context around to control the calling chain timeout

Hexilee:
Split from #409

Currently, the top context isn't passed around, which may cause dead goroutine in kubelet or controller-runtime.

### DIFF
--- a/pkg/controller/reconciler.go
+++ b/pkg/controller/reconciler.go
@@ -17,16 +17,16 @@ limitations under the License.
 package controller
 
 import (
+	"context"
 	"strconv"
 	"time"
 
 	"github.com/juicedata/juicefs-csi-driver/pkg/juicefs/k8sclient"
-	"golang.org/x/net/context"
+	"k8s.io/klog"
+	k8sexec "k8s.io/utils/exec"
 	"k8s.io/utils/mount"
 
 	"github.com/juicedata/juicefs-csi-driver/pkg/config"
-	"k8s.io/klog"
-	k8sexec "k8s.io/utils/exec"
 )
 
 type PodReconciler struct {
--- a/pkg/controller/reconciler.go
+++ b/pkg/controller/reconciler.go
@@ -85,7 +85,7 @@ func doReconcile(kc *kubeletClient, driver *PodDriver) {
 			if value, ok := pod.Labels[config.PodTypeKey]; !ok || value != config.PodTypeValue {
 				continue
 			}
-			ctx, cancel := context.WithTimeout(context.Background(), config.ContextTimeout)
+			ctx, cancel := context.WithTimeout(context.Background(), config.ReconcileTimeout)
 			defer cancel()
 			err := driver.Run(ctx, pod)
 			if err != nil {
--- a/pkg/controller/reconciler.go
+++ b/pkg/controller/reconciler.go
@@ -17,11 +17,11 @@ limitations under the License.
 package controller
 
 import (
-	"context"
 	"strconv"
 	"time"
 
 	"github.com/juicedata/juicefs-csi-driver/pkg/juicefs/k8sclient"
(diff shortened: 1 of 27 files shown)


pkg/controller/reconciler.go
```diff
@@ -17,11 +17,11 @@ limitations under the License.
 package controller
 
 import (
-	"context"
 	"strconv"
 	"time"
 
 	"github.com/juicedata/juicefs-csi-driver/pkg/juicefs/k8sclient"
+	"golang.org/x/net/context"
```
davies: ？

pkg/juicefs/juicefs.go
```diff
@@ -32,6 +31,7 @@ import (
 	"time"
 
 	"github.com/juicedata/juicefs-csi-driver/pkg/util"
+	"github.com/pkg/errors"
```
davies, line 34: ？

pkg/juicefs/mount/pod_mount.go
```diff
@@ -348,23 +347,26 @@ func (p *PodMount) waitUtilMountReady(jfsSetting *jfsConfig.JfsSetting, podName
...
-	for i := 0; i < 120; i++ {
-		job, err := p.K8sClient.GetJob(jobName, jfsConfig.Namespace)
+	for {
+		job, err := p.K8sClient.GetJob(ctx, jobName, jfsConfig.Namespace)
 		if err != nil {
 			if k8serrors.IsNotFound(err) {
 				klog.Infof("waitUtilJobCompleted: Job %s is completed and been recycled", jobName)
 				return nil
 			}
+			if k8serrors.IsTimeout(err) {
```
davies: ctx may not have timeout specified, so this will change the behavior (not expected).

pkg/juicefs/juicefs.go
```diff
@@ -687,15 +684,11 @@ func (j *juicefs) ceFormat(secrets map[string]string, noUpdate bool, setting *co
 		envs = append(envs, "JFS_NO_CHECK_OBJECT_STORAGE=1")
 	}
 	formatCmd.SetEnv(envs)
-	var res []byte
-	_, err := util.DoWithinTime(context.TODO(), 50*time.Second, nil, func() (err error) {
```
davies, line 691: We don't add timeout for `ctx`, let's keep this for this PR.

davies: We should keep these change in #409

pkg/juicefs/mount/pod_mount.go
Hexilee: Changes reverted

pkg/controller/reconciler.go
Hexilee: Changed to stdlib

pkg/juicefs/juicefs.go
Hexilee: The `github.com/pkg/errors` has more features than `errors` like [`Wrap`](https://pkg.go.dev/github.com/pkg/errors#Wrap) function.

pkg/juicefs/juicefs.go
Hexilee: Changes reverted

pkg/juicefs/mount/pod_mount.go
```diff
@@ -288,9 +287,9 @@ func (p *PodMount) createOrAddRef(jfsSetting *jfsConfig.JfsSetting) (podName str
 	r := builder.NewBuilder(jfsSetting)
 	secret := r.NewSecret()
 	key := util.GetReferenceKey(jfsSetting.TargetPath)
-	for i := 0; i < 120; i++ {
+	for {
```
davies: This is an infinity loop if no timeout in context.

davies (approved):

PR #415, opened 2022-09-02
Fuse image

zwwhdls:
make fuse image for Fluid

### DIFF
--- a/fuse.Dockerfile
+++ b/fuse.Dockerfile
@@ -1,0 +1,40 @@
+FROM golang:1.16-buster as builder
+
+ARG GOPROXY
+ARG JUICEFS_REPO_BRANCH=main
+ARG JUICEFS_REPO_REF=${JUICEFS_REPO_BRANCH}
+
+WORKDIR /workspace
+ENV GOPROXY=${GOPROXY:-https://proxy.golang.org}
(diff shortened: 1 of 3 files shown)


fuse.Dockerfile
```diff
@@ -0,0 +1,40 @@
...
+
+WORKDIR /app
+
+ENV JUICEFS_CLI=/usr/bin/juicefs
+ENV JFS_AUTO_UPGRADE=${JFS_AUTO_UPGRADE:-enabled}
+ENV JFS_MOUNT_PATH=/usr/local/juicefs/mount/jfsmount
+
+RUN apt-get update && apt-get install -y librados2 curl fuse && \
+    rm -rf /var/cache/apt/* && \
+    curl -sSL https://juicefs.com/static/juicefs -o ${JUICEFS_CLI} && chmod +x ${JUICEFS_CLI} && \
```
Hexilee, line 26: We should also specify `TARGETARCH` here

Hexilee (approved):
Rest LGTM

fuse.Dockerfile
Hexilee: Oops, it's unnecessary.

PR #423, opened 2022-09-08
add mount share & pathpattern test in ci

zwwhdls:


.github/scripts/test_case.py
```diff
@@ -1521,39 +937,230 @@ def test_deployment_static_patch_pv():
...
+        raise Exception("Pods of deployment {} are not delete within 5 min.".format(deployment.name))
+    LOG.info("Remove pvc {}".format(pvc.name))
+    pvc.delete()
+    LOG.info("Remove pv {}".format(pvc.name))
+    pv.delete()
+    LOG.info("Test pass.")
+    return
+
+
+def test_share_mount():
```
Hexilee, line 1058: Shall we set `pathPattern` here to cover #413?

Hexilee (approved):
Rest LGTM

.github/scripts/test_case.py
zwwhdls: `pathPattern` is not the root cause in this case. `subpath` is different when using different PV and this test case has covered it.

PR #428, opened 2022-09-16
update log & fix ctx exceeded & add ctx timeout in getting uuid

zwwhdls:
1. update log
2. fix ctx exceeded 
3. add ctx timeout in getting uuid
4. update timeout of job to 40s because parent ctx timeout is 60s


Hexilee (approved):
LGTM

Issue #354, opened 2022-06-17
[FEATURE] create node pod on demand


davies:
Right now, we use node pod as daemonset, so it's deployed into every node even it's not used by any app. 

We could create the node pod on demand by controller pod.

zwwhdls:
Plan:
1. Add nodeSelector to CSI node daemonset, similar to: juicefs-csi-node: true
2. Enable CSI controller attach
	1. Implement the ControllerPublishVolume interface of controller
	2. csiDriver shows the specified attach=required
	3. Add csi-attacher sidecar to csi controller
3. Check whether the corresponding node has csi node pod in ControllerPublishVolume interface:
	1. If not, add the corresponding label to the node
	2. Wait for csi node pod to start
4. No operation in the ControllerUnpublishVolume interface


Potential Problem:
1. CSI can't upgrade smoothly
	1. Existing csi node pods will be deleted after upgrade
	2. csiDriver resources cannot be upgraded, only deleted and rebuilt
2. Attach sidecar is not compatible with K8s version
	1. k8s v0.14-v0.17: quay.io/k8scsi/csi-attacher:v2.2.0
	2. k8s v0.17+: k8s.gcr.io/sig-storage/csi-attacher:v3.4.0
3. It will spend lots of time of creating PV for the first time.
	The first time you create a PV, you need to wait for the csi node pod to start, which involves pulling images and other processes, and the startup time is also affected by the cluster load
4. When uninstalling csi, the label on the node needs to be deleted manually

fishincat:
make this function choosable?

zwwhdls:
@fishincat Of course, it will be off by default.

zwwhdls:
The plan above may result in slow PVC mount. It's because the `csi-attacher` need to watch `volumeAttach` and call csi controller and then set `volumeAttach` completed. It spent some watch period before volume mount. 

Therefore, we temporarily abandon this solution and add a document to explain how users can manually implement the on-demand startup of CSI node.

PR #429, opened 2022-09-27
setup csi node on demand in controller 

zwwhdls:
1. setup csi node on demand in controller 
2. add attacher sidecar in controler 
3. update attachRequired true


zwwhdls:
close it. refer to https://github.com/juicedata/juicefs-csi-driver/issues/354#issuecomment-1305009465

PR #435, opened 2022-10-21
fix subpath not deleted when using path pattern and add delete pvc test in pathpattern

### LINKED ISSUE #436: [BUG] subpath not deleted when using path pattern
**What happened**:

**Environment**:
- Others:

zwwhdls:
fix https://github.com/juicedata/juicefs-csi-driver/issues/436 
and add ci test of it


Hexilee (approved):
LGTM

PR #454, opened 2022-11-10
support juicefs client as sidecar in application pod

### LINKED ISSUE #448: [FEATURE] support juicefs client as sidecar in application pod

zwwhdls:
close https://github.com/juicedata/juicefs-csi-driver/issues/448


zwwhdls:
I will split this PR into some small PRs and keep main feature more clear.

zwwhdls:
after https://github.com/juicedata/juicefs-csi-driver/pull/458 and https://github.com/juicedata/juicefs-csi-driver/pull/461

pkg/webhook/handler/mutate/sidecar.go
```diff
@@ -0,0 +1,197 @@
...
+		PVC:     pvc,
+		PV:      pv,
+	}
+}
+
+func (s *SidecarMutate) Mutate(pod *corev1.Pod) (out *corev1.Pod, err error) {
+	// get secret, volumeContext and mountOptions from PV
+	secrets, volCtx, options, err := s.GetSettings(*s.PV)
+	if err != nil {
+		klog.Infof("get settings from pv %s of pod %s namespace %s err: %v", s.PV.Name, pod.Name, pod.Namespace, err)
```
Hexilee: Should it be `klog.Errorf` or just wrap this error and return?

pkg/webhook/handler/mutate/sidecar.go
```diff
@@ -0,0 +1,197 @@
...
+func NewSidecarMutate(client *k8sclient.K8sClient, jfs juicefs.Interface, pvc *corev1.PersistentVolumeClaim, pv *corev1.PersistentVolume) Mutate {
+	return &SidecarMutate{
+		Client:  client,
+		juicefs: jfs,
+		PVC:     pvc,
+		PV:      pv,
+	}
+}
+
+func (s *SidecarMutate) Mutate(pod *corev1.Pod) (out *corev1.Pod, err error) {
```
Hexilee: I think this method should pass context from caller, as webhook should timeout fastly, and the s.createOrUpdateSecret may cost much time.

pkg/webhook/handler/handler.go
```diff
@@ -0,0 +1,148 @@
...
+}
+
+// InjectDecoder injects the decoder.
+func (s *SidecarHandler) InjectDecoder(d *admission.Decoder) error {
+	s.decoder = d
+	return nil
+}
+
+// GetVolume get juicefs pv & pvc from pod
+func (s *SidecarHandler) GetVolume(pod corev1.Pod) (used bool, pvGot *corev1.PersistentVolume, pvcGot *corev1.PersistentVolumeClaim, err error) {
```
Hexilee: Also need context from the caller.

Hexilee (approved):
Almost LGTM, good job! There left two problems:
1. The origin containers should wait the `jfs-mount` container, resulting in the `jfs-mount` must be the first (default) container. For example, the `kubectl logs` or `kubectl exec` will select the `jfs-mount` by default instead of the origin containers.
2. The pod cannot complete as expected, because the `jfs-mount` container won't exit. We need watch the injected pods to complete them.

PR #458, opened 2022-11-14
separate cmd flag of controller and node

zwwhdls:


cmd/node.go
```diff
@@ -0,0 +1,146 @@
...
+		for {
+			http.ListenAndServe(fmt.Sprintf("localhost:%d", port), nil)
+			port++
+		}
+	}()
+
+	// enable pod manager in csi node
+	if !process && podManager && config.KubeletPort != "" && config.HostIp != "" {
+		if err := controller.StartReconciler(); err != nil {
+			klog.V(5).Infof("Could not Start Reconciler: %v", err)
```
Hexilee: I think `klog.Fatalln` is better here.

Hexilee (approved):
LGTM

cmd/node.go
zwwhdls: done

zwwhdls:
break change, closing

PR #459, opened 2022-11-14
fix manage mount pod in csi controller when csi node is gone

### LINKED ISSUE #460: [BUG] when csi node is gone and csi controller restarts, remaining mount pod stick in terminating
**What happened**:
When csi node and controller are all deleted, and then node is scaled down, the remaining mount pod will be sticked in Terminating. 

**Environment**:
- Others:

zwwhdls:
close https://github.com/juicedata/juicefs-csi-driver/issues/460

An old [PR](https://github.com/juicedata/juicefs-csi-driver/pull/347) solved this issue. But controller only watches update & delete event of mount pod. when controller restart, it can't receive create event and cause leak of mount pod. So it should watch create event also.


Hexilee (approved):
LGTM

PR #473, opened 2022-11-22
fix deadlock and make mount pod reconcile in goroutine

### LINKED ISSUE #467: [BUG] CSI node repeatedly checks the mountpoint and falls into an infinite loop
**What happened**:
When mount pod did not ready, CSI Node repeatedly checks the mountpoint and falls into an infinite loop. 

Lots of duplicate logs:
```
I1118 03:56:28.812876       1 pod_mount.go:346] mount path /jfs/pv-6-oxzqll not ready: stat /jfs/pv-6-oxzqll: no such file or directory
```

CSI Node should return error and wait for kubelet retry.

**Environment**:
- JuiceFS CSI Driver version (which image tag did your CSI Driver use): v0.17.2
 ```
Client Version: version.Info{Major:"1", Minor:"23", GitVersion:"v1.23.1", GitCommit:"86ec240af8cbd1b60bcc4c03c20da9b98005b92e", GitTreeState:"clean", BuildDate:"2021-12-16T11:41:01Z", GoVersion:"go1.17.5", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"23", GitVersion:"v1.23.3", GitCommit:"816c97ab8cff8a1c72eccca1026f7820e93e0d25", GitTreeState:"clean", BuildDate:"2022-01-25T21:19:12Z", GoVersion:"go1.17.6", Compiler:"gc", Platform:"linux/amd64"}
```

- Others:

### LINKED ISSUE #472: [BUG] Mount failed when hundreds of pods mounted different PVC
**What happened**:
When hundreds of pods mounted different PVCs created at the same time, many mount pods failed because of `UnexpectedAdmissionError`. CSI Node service will delete resources of mount pod and redeploy it when node resources are in short supply. But only a few mount pod redeployed.

**Environment**:
- Others:

### PR DESCRIPTION
**Pull request by zwwhdls**
fix deadlock when mount pod faild. fix https://github.com/juicedata/juicefs-csi-driver/issues/467
make mount pod reconcile in goroutine. fix https://github.com/juicedata/juicefs-csi-driver/issues/472

### DIFF
--- a/pkg/controller/pod_driver.go
+++ b/pkg/controller/pod_driver.go
@@ -41,7 +41,7 @@ const defaultCheckoutTimeout = 1 * time.Second
 type PodDriver struct {
 	Client   *k8sclient.K8sClient
 	handlers map[podStatus]podHandler
-	mit      *mountInfoTable
+	mit      mountInfoTable
 	mount.SafeFormatAndMount
 }
 
@@ -53,7 +53,6 @@ func newPodDriver(client *k8sclient.K8sClient, mounter mount.SafeFormatAndMount)
 	driver := &PodDriver{
 		Client:             client,
 		handlers:           map[podStatus]podHandler{},
-		mit:                newMountInfoTable(),
 		SafeFormatAndMount: mounter,
 	}
 	driver.handlers[podReady] = driver.podReadyHandler
@@ -72,6 +71,10 @@ const (
 	podDeleted podStatus = "podDeleted"
 	podPending podStatus = "podPending"
 )
+
+func (p *PodDriver) SetMountInfo(mit mountInfoTable) {
+	p.mit = mit
+}
 
 func (p *PodDriver) Run(ctx context.Context, current *corev1.Pod) error {
 	// check refs in mount pod annotation first, delete ref that target pod is not found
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: pkg/controller/pod_driver.go [hunk 1]
```diff
@@ -188,7 +191,7 @@ func (p *PodDriver) podErrorHandler(ctx context.Context, pod *corev1.Pod) error
 
 	// check resource err
 	if util.IsPodResourceError(pod) {
-		klog.V(5).Infof("waitUtilMount: Pod is failed because of resource.")
+		klog.V(5).Infof("waitUtilMount: Pod %s is failed because of resource.", pod.Name)
```
davies: no 'is'

pkg/controller/reconciler.go
```diff
@@ -85,16 +80,27 @@ func doReconcile(kc *kubeletClient, driver *PodDriver) {
 			if value, ok := pod.Labels[config.PodTypeKey]; !ok || value != config.PodTypeValue {
 				continue
 			}
-			err := func() error {
+			wg.Add(1)
+			go func() {
+				defer wg.Done()
```
Hexilee: errgroup should be more suitable for this case.

pkg/controller/reconciler.go
```diff
@@ -85,16 +80,27 @@ func doReconcile(kc *kubeletClient, driver *PodDriver) {
...
 				ctx, cancel := context.WithTimeout(context.Background(), config.ReconcileTimeout)
 				defer cancel()
-				return driver.Run(ctx, pod)
+				mounter := mount.SafeFormatAndMount{
+					Interface: mount.New(""),
+					Exec:      k8sexec.New(),
+				}
+
+				driver := NewPodDriver(ks, mounter)
```
Hexilee: Why do we need an independent driver for each Pod?

zwwhdls: podDriver handlers is a map and select a handler by pod status. If all pods goroutine use one driver may result in `concurrent map writes map`. Any better idea?

Hexilee: Ok, looks good to me.

pkg/controller/reconciler.go
zwwhdls: PTAL

PR #474, opened 2022-11-22
make create subpath as postStart hook in mount pod instead of mount & create in CSI Controller

### LINKED ISSUE #471: [FEATURE] config create subpath as postStart hook in mount pod instead of mount & create in CSI Controller
CSI Controller will mount and create subpath in juicefs per PVC. When hundreds of PVCs are created at the same time, lots of mount process/pod will be created, resulting in CSI Controller hang. And when delete CSI Controller, main process and its subprocesses are potential to turned into zombies with docker lower version. 

zwwhdls:
fix https://github.com/juicedata/juicefs-csi-driver/issues/471

### DIFF
--- a/pkg/driver/controller_test.go
+++ b/pkg/driver/controller_test.go
@@ -91,7 +91,6 @@ func TestCreateVolume(t *testing.T) {
 				mockCtl := gomock.NewController(t)
 				defer mockCtl.Finish()
 				mockJuicefs := mocks.NewMockInterface(mockCtl)
-				//mockJuicefs.EXPECT().JfsCreateVol(context.TODO(), volumeId, volumeId, secret, volCtx).Return(nil)
 
 				juicefsDriver := controllerService{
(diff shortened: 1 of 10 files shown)


pkg/driver/controller_test.go
```diff
@@ -91,7 +91,7 @@ func TestCreateVolume(t *testing.T) {
 				mockCtl := gomock.NewController(t)
 				defer mockCtl.Finish()
 				mockJuicefs := mocks.NewMockInterface(mockCtl)
-				mockJuicefs.EXPECT().JfsCreateVol(context.TODO(), volumeId, volumeId, secret, volCtx).Return(nil)
+				//mockJuicefs.EXPECT().JfsCreateVol(context.TODO(), volumeId, volumeId, secret, volCtx).Return(nil)
```
Hexilee: May we should remove this line?

Hexilee (approved):
Rest LGTM

PR #483, opened 2022-11-28
fallback to accessing apiserver when being denied by kubelet

### LINKED ISSUE #478: [BUG] fallback to accessing apiserver when being denied by kubelet
**What happened**:

csi-node does not work, and print logs:

```log
I1124 07:35:19.434296       6 kubelet_client.go:99] GetNodeRunningPods err: Unauthorized
E1124 07:35:19.434334       6 reconciler.go:70] doReconcile GetNodeRunningPods: invalid character 'U' looking for beginning of value
```

Everything's ok

Disable webhook authentication in kubelet config, then depoly juicefs-csi-driver with default configuration.

We support rbac access to kubelet in https://github.com/juicedata/juicefs-csi-driver/pull/287, but we still fail to access kubelet when it disabled webhook authentication. We should list pods by apiserver when kubelet denied our requests.

**Environment**:
- Others:

Hexilee:
close #478

### DIFF
--- a/pkg/controller/reconciler.go
+++ b/pkg/controller/reconciler.go
@@ -66,8 +66,16 @@ func doReconcile(ks *k8sclient.K8sClient, kc *kubeletClient) {
 		mit := newMountInfoTable()
 		podList, err := kc.GetNodeRunningPods()
 		if err != nil {
-			klog.Errorf("doReconcile GetNodeRunningPods: %v", err)
-			goto finish
+			klog.Warningf(`failed to GetNodeRunningPods from kubelet("%s"), fallback to apiserver`, err.Error())
+			if config.CSIPod.Spec.NodeName != "" {
+				podList, err = ks.ListPodsOnNode(ctx, config.CSIPod.Spec.NodeName)
+			} else {
+				podList, err = ks.ListPodsOnNodeIP(ctx, config.CSIPod.Status.HostIP)
+			}
+			if err != nil {
+				klog.Errorf("failed to ListPodsOnNode: %v", err)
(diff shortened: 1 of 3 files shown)


pkg/controller/reconciler.go
```diff
@@ -66,8 +66,16 @@ func doReconcile(ks *k8sclient.K8sClient, kc *kubeletClient) {
 		mit := newMountInfoTable()
 		podList, err := kc.GetNodeRunningPods()
 		if err != nil {
-			klog.Errorf("doReconcile GetNodeRunningPods: %v", err)
-			goto finish
+			klog.Warningf(`failed to GetNodeRunningPods from kubelet("%s"), fallback to apiserver`, err.Error())
+			if config.CSIPod.Spec.NodeName != "" {
+				podList, err = ks.ListPodsOnNode(ctx, config.CSIPod.Spec.NodeName)
+			} else {
+				podList, err = ks.ListPodsOnNodeIP(ctx, config.CSIPod.Status.HostIP)
```
zwwhdls, line 73: Polling Apiserver by list pod will cause a lot of pressure on Apiserver, it is recommended to change it to watch.

zwwhdls (changes requested):

pkg/controller/reconciler.go
Hexilee: As we need the target pods of mount pods, we cannot avoid listing `Pod` on the `apiserver`.

zwwhdls:
Still not find a way to avoid heavy stress to apiserver. closing.

PR #492, opened 2022-12-05
add offline.md

timfeirg:
copy images to an offline cluster


timfeirg:
will add translation after review complete

PR #501, opened 2022-12-07
make name field more easier to find in k8s.yaml

timfeirg:
name is hidden all the way down inside all those YAMLs, finding desired
container is really hard.

### DIFF
--- a/deploy/k8s.yaml
+++ b/deploy/k8s.yaml
@@ -241,7 +241,8 @@ spec:
         app.kubernetes.io/version: master
     spec:
       containers:
-      - args:
+      - name: juicefs-plugin
+        args:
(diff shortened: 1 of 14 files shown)


zwwhdls:
deploy/k8s.yaml and deploy/k8s_before_v1_18.yaml are generated by kustomize automatically. Please do not modify them directly. @timfeirg

timfeirg:
I'll look into kustomize configs to see if we can control its print order. `k8s.yaml` is a file that often requires manual editing, I think it's worth polishing.

timfeirg:
no easy way to accomplish this, abort.

> I'll look into kustomize configs to see if we can control its print order.

PR #522, opened 2022-12-20
check resource version when del mount pod

zwwhdls:


Hexilee (approved):
LGTM

pkg/controller/pod_driver.go
```diff
@@ -133,26 +127,36 @@ func (p *PodDriver) checkAnnotations(ctx context.Context, pod *corev1.Pod) error
...
-	if len(pod.Annotations) != len(annotation) {
-		if err := util.PatchPodAnnotation(ctx, p.Client, pod, annotation); err != nil {
-			klog.Errorf("Update pod %s error: %v", pod.Name, err)
+	if len(delAnnotations) != 0 {
+		// check mount pod resourceVersion, if it is not the latest, return conflict
+		newPod, err := p.Client.GetPod(ctx, pod.Name, pod.Namespace)
+		if err != nil {
+			return err
+		}
+		if newPod.ResourceVersion != pod.ResourceVersion {
```
Hexilee, line 153: I think this version check is unnecessary, as the following `DelPodAnnotation` should return a conflict error if the pod is outdated

Hexilee (approved):
Rest LGTM

pkg/controller/pod_driver.go
zwwhdls: patch api will not check resourceVersion, so we check it before patch.

Issue #529, opened 2022-12-26
[FEATURE] Add port for prometheus monitoring to mount pod definition


dbotwinick:
**What feature you'd like to add:**
Add port for prometheus monitoring to juicefs-mount pod definition(s)

Would just require a small addition for the containers spec...

```
    ports:
      - containerPort: 9567
        name: prometheus

```
**Why is this feature needed:**
Would facilitate easy prometheus monitoring via PodMonitor CRD from the prometheus-operator

Example PodMonitor definition...

```
apiVersion: monitoring.coreos.com/v1
kind: PodMonitor
metadata:
  name: juicefs-mounts-monitor
  namespace: monitoring
  labels:
    name: juicefs-mounts-monitor
spec:
  namespaceSelector:
    matchNames:
      - kube-system
  selector:
    matchLabels:
      app.kubernetes.io/name: juicefs-mount
  podMetricsEndpoints:
    - targetPort: prometheus
      path: '/metrics'
      scheme: 'http'
      interval: '5s'
```
**Other Considerations**
Would require updating port if the port is otherwise configured and would require disabling port if monitoring features are disabled.

robgordon89:
@dbotwinick this was reworked and released a few days ago and works nicely on my cluster 👍 

The only thing that changed was the port name.

```
apiVersion: monitoring.coreos.com/v1
kind: PodMonitor
metadata:
  name: juicefs-mounts-monitor
  namespace: monitoring
  labels:
    name: juicefs-mounts-monitor
spec:
  namespaceSelector:
    matchNames:
      - kube-system
  selector:
    matchLabels:
      app.kubernetes.io/name: juicefs-mount
  podMetricsEndpoints:
    - targetPort: metrics
      path: "/metrics"
      scheme: "http"
      interval: "5s"
```

zwwhdls:
closed by https://github.com/juicedata/juicefs-csi-driver/pull/552

PR #530, opened 2022-12-26
hacked in naïve approach to adding ports to mount pods that assumes p…

dbotwinick:
…ort will always be enabled and is hardcoded to 9567 (matches default configuration).

This PR is a quick/dirty implementation starter for the feature request: https://github.com/juicedata/juicefs-csi-driver/issues/529

...and is basically just intended to reduce the barrier to implementation rather than be a solid, verified implementation.

### DIFF
--- a/pkg/juicefs/mount/builder/pod.go
+++ b/pkg/juicefs/mount/builder/pod.go
@@ -56,6 +56,9 @@ func (r *Builder) NewMountPod(podName string) *corev1.Pod {
 				"umount %s && rmdir %s", r.jfsSetting.MountPath, r.jfsSetting.MountPath)}},
 		},
 	}
+	pod.Spec.Containers[0].Ports = []corev1.ContainerPort{
+		{Name: "prometheus", ContainerPort: 9567},
+	}
 	gracePeriod := int64(10)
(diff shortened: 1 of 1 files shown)


zwwhdls:
Hi @dbotwinick , thanks for your contribution! `metrics` can be set by mountOptions `metrics=`, can you obtain the metrics port from this option and make `9567` as default instead of hardcoding?

dbotwinick:
Hi @zwwhdls. I'll try to do it more properly later. I honestly just had a few moments and thought it'd be nice to do _something_ other than _just_ requesting a feature. I'll circle back around to it and make it compatible with setting metrics via mountOptions. I guess I should also update the associated tests!

zwwhdls:
Thanks very much again! @dbotwinick You can kindly get some references [here](https://github.com/juicedata/juicefs-csi-driver/blob/master/pkg/juicefs/mount/builder/pod.go#L131)

robgordon89:
i had a stab at this as this has been a month 

https://github.com/juicedata/juicefs-csi-driver/pull/552

Still not completely happy with using regex to extract the port we need but its better.

zwwhdls:
dup to https://github.com/juicedata/juicefs-csi-driver/pull/552. closing

Issue #531, opened 2022-12-27
[BUG] csi controller crashloopbackoff


timfeirg:
**What happened**: 

events:

```
Events:
  Type     Reason     Age                 From               Message
  ----     ------     ----                ----               -------
  Normal   Scheduled  119s                default-scheduler  Successfully assigned kube-system/juicefs-csi-controller-0 to k8s-worker-01
  Normal   Killing    68s                 kubelet            Container juicefs-plugin failed liveness probe, will be restarted
  Normal   Pulled     38s (x2 over 118s)  kubelet            Container image "juicedata/juicefs-csi-driver:v0.17.3" already present on machine
  Normal   Created    38s (x2 over 118s)  kubelet            Created container juicefs-plugin
  Normal   Started    38s (x2 over 118s)  kubelet            Started container juicefs-plugin
  Warning  Unhealthy  8s (x8 over 108s)   kubelet            Liveness probe failed: Get "http://10.142.0.232:9909/healthz": dial tcp [10.142.0.232:9909](http://10.142.0.232:9909/): connect: connection refused
```

csi controller log:

```
I1227 05:53:26.654089       7 client.go:89] Get pod juicefs-csi-controller-0
I1227 05:53:26.663498       7 driver.go:44] Driver: csi.juicefs.com version v0.17.3 commit b80e312364faea35180e5c3de8b9f696cbe1b36d date 2022-11-24T11:36:27Z
I1227 05:53:26.665258       7 driver.go:109] Listening for connection on address: &net.UnixAddr{Name:"/var/lib/csi/sockets/pluginproxy/csi.sock", Net:"unix"}
I1227 05:53:26.666332       7 leaderelection.go:248] attempting to acquire leader lease kube-system/csi.juicefs.com...
I1227 05:53:27.714258       7 request.go:665] Waited for 1.047909251s due to client-side throttling, not priority and fairness, request: GET:https://10.96.0.1:443/apis/serving.kserve.io/v1beta1?timeout=32s
I1227 05:53:28.466724       7 main.go:154] Mount Manager Started
...
I1227 05:53:43.736812       7 leaderelection.go:258] successfully acquired lease kube-system/csi.juicefs.com
I1227 05:53:43.736981       7 controller.go:835] Starting provisioner controller csi.juicefs.com_juicefs-csi-controller-0_2b40d83e-5b97-4ca6-b81e-164f9a7460c6!
[...]

PR #532, opened 2022-12-27
move livenessProbe to where it belongs

timfeirg:
suggestion: could we delete liveness probe container altogether? since it doesn't really represent the liveness & healthness of the csi controller plugin container.

related: https://github.com/juicedata/juicefs-csi-driver/issues/531


zwwhdls:
As shows in https://github.com/kubernetes-csi/livenessprobe/blob/master/README.md#usage, actual `livenessProbe` is set on the container with the CSI driver. closing.

PR #552, opened 2023-01-19
chore: add metrics port to mount pod

robgordon89:
Adding metrics port to mount pod, the approach here I'm not entirely happy with the getMetricsPort function, had to do some regex on the options 🤔 not sure how else to tackle it, unless it is moved to its own config.

Added tests to show that it is overridden when you update the metric port in mount options.

Set 9567 as default so it wont break current deployments


pkg/juicefs/mount/builder/pod.go
```diff
@@ -194,3 +204,20 @@ func (r *Builder) getInitContainer() corev1.Container {
 	})
 	return container
 }
+
+func (r *Builder) getMetricsPort() int32 {
```
zwwhdls, line 217: Can you add unit test for this function?

zwwhdls (changes requested):

zwwhdls:
Can you also fix the code scanning alert? Thanks. https://github.com/juicedata/juicefs-csi-driver/security/code-scanning/2

pkg/juicefs/mount/builder/pod.go
robgordon89: Sorry for the delay on this 

unit test added

zwwhdls:
Thanks for your contribution! @robgordon89

PR #610, opened 2023-03-24
set mount pod unprivileged

zwwhdls:


zwwhdls:
After deep research, `privileged` is required for mount pod to propagate mountpoint back to the host  , referring to https://kubernetes.io/docs/concepts/storage/volumes/#mount-propagation. 

closing.

PR #637, opened 2023-04-10
add ha docs && add npm check md in makefile

zwwhdls:


timfeirg:
in the latest commit, I've:

* move HA related docs to the bottom, considering HA is enabled by default, and need little attention
* move helm part to the front, since helm is more recommended
* further simplified language

PR #645, opened 2023-04-24
fix if pod namespace is empty && fix pod using multiple juicefs pvc

### LINKED ISSUE #644: [BUG] sidecar can not be injected in CCE
**What happened**:
When create a deployment in CCE, pod can not be created , error message in controller:

```
Can't get pvc ce-static in namespace  : an empty namespace may not be set when a resource name is provided
```

**Environment**:
- Others:

zwwhdls:
close https://github.com/juicedata/juicefs-csi-driver/issues/644


pkg/webhook/handler/config/volume.go
```diff
@@ -0,0 +1,179 @@
...
+				// if storageclass is juicefs
+				if sc.Provisioner == config.DriverName {
+					used = true
+				}
+			}
+
+			// get PV
+			var pv *corev1.PersistentVolume
+			if pvc.Spec.VolumeName == "" {
+				if used {
```
Hexilee, line 85: should we clear `used` for each volume?

Hexilee (approved):
Rest LGTM

pkg/webhook/handler/config/volume.go
zwwhdls: my fault, fix it.

PR #648, opened 2023-05-08
fix delvol image

### LINKED ISSUE #642: [BUG] delvol image should inherit mount-image
**What happened**: `juicefs/mount-image` is already set, but delvol pod is still using the default container

zwwhdls:
fix https://github.com/juicedata/juicefs-csi-driver/issues/642


Hexilee (approved):
LGTM

PR #661, opened 2023-06-06
recycle job when ttl not work

### LINKED ISSUE #659: [BUG] clean cache job will not be recycled when TTLAfterFinished is not enabled
**What happened**:

**Environment**:
- Others:

zwwhdls:
fix https://github.com/juicedata/juicefs-csi-driver/issues/659

### DIFF
--- a/pkg/juicefs/mount/pod_mount.go
+++ b/pkg/juicefs/mount/pod_mount.go
@@ -400,7 +400,14 @@ func (p *PodMount) waitUtilJobCompleted(ctx context.Context, jobName string) err
 		}
 		if util.IsJobCompleted(job) {
 			klog.V(5).Infof("waitUtilJobCompleted: Job %s is completed", jobName)
-			return nil
+			if util.IsJobShouldBeRecycled(job) {
+				// try to delete job
+				klog.Infof("job %s completed but not be recycled automatically, delete it", jobName)
+				if err := p.K8sClient.DeleteJob(ctx, jobName, jfsConfig.Namespace); err != nil {
+					klog.Errorf("delete job %s error %v", jobName, err)
+					return err
+				}
+			}
 		}
 		time.Sleep(time.Millisecond * 500)
 	}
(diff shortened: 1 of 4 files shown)


pkg/juicefs/mount/pod_mount.go
```diff
@@ -400,7 +400,14 @@ func (p *PodMount) waitUtilJobCompleted(ctx context.Context, jobName string) err
 		}
 		if util.IsJobCompleted(job) {
 			klog.V(5).Infof("waitUtilJobCompleted: Job %s is completed", jobName)
-			return nil
+			if util.IsJobShouldBeRecycled(job) {
+				// try to delete job
+				klog.Infof("job %s completed but not be recycled automatically, delete it", jobName)
+				if err := p.K8sClient.DeleteJob(ctx, jobName, jfsConfig.Namespace); err != nil {
+					klog.Errorf("delete job %s error %v", jobName, err)
+					return err
```
Hexilee: This `err` should always be returned even if it's nil, otherwise, there will be a dead loop.

pkg/juicefs/mount/pod_mount.go
```diff
@@ -410,7 +417,7 @@ func (p *PodMount) waitUtilJobCompleted(ctx context.Context, jobName string) err
 			"job-name": jobName,
 		},
 	}, nil)
-	if err != nil || len(pods) != 1 {
+	if err != nil || len(pods) == 0 {
```
Hexilee, line 420: maybe `len(pods) != 0`?

zwwhdls: len(pods) == 0 means no pod of job is created, it's abnormal and need to return error.

Hexilee: Ok

pkg/juicefs/mount/pod_mount.go
zwwhdls: done

Hexilee (approved):

PR #680, opened 2023-06-28
Introduce `JUICEFS_IMMUTABLE`, mount `/etc/updatedb.conf` only in mutable environments

twelho:
Some Kubernetes distributions, such as [Talos Linux](https://www.talos.dev/), do not ship `updatedb` at all. In these environments, the JuiceFS CSI mount pod fails to start due to `/etc/updatedb.conf` not being present on the host for bind mounting, nor can `FileOrCreate` create it due to `/etc` being read-only. This patch introduces `JUICEFS_IMMUTABLE` as a new environment variable for both controller and node instances to handle these kinds of cases that need special attention in immutable environments. I've also updated `updatedb`-related hard-coded strings to use the constant definitions in `common.go`.

### DIFF
--- a/cmd/controller.go
+++ b/cmd/controller.go
@@ -21,6 +21,7 @@ import (
 	"fmt"
 	"net/http"
 	"os"
+	"strconv"
 
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/runtime"
@@ -62,6 +63,14 @@ func parseControllerConfig() {
 		config.FormatInPod = true
 		config.MountManager = false
 		config.ByProcess = false
+	}
+	if jfsImmutable := os.Getenv("JUICEFS_IMMUTABLE"); jfsImmutable != "" {
+		// check if running in an immutable environment
+		if immutable, err := strconv.ParseBool(jfsImmutable); err == nil {
+			config.Immutable = immutable
+		} else {
+			klog.Errorf("cannot parse JUICEFS_IMMUTABLE: %v", err)
+		}
 	}
 
 	config.NodeName = os.Getenv("NODE_NAME")
--- a/cmd/node.go
+++ b/cmd/node.go
@@ -21,6 +21,7 @@ import (
 	"fmt"
 	"net/http"
 	"os"
+	"strconv"
 	"time"
 
 	"k8s.io/klog"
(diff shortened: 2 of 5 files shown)


zwwhdls:
Hi @twelho , CSI Node does not mount `/etc/updatedb.conf` inside, so CSI Node pod can not stat it inside container.

We alreay add `FileOrCreate` in `volume.hostPath`, I think it doesn't matter.

twelho:
Hmm, you're right, `os.Stat` inside the container won't work. However, since the filesystem in Talos is immutable, `/etc` is read-only, which means that `FileOrCreate` will not be able to create the file, leading to the failure I'm observing. Would it be possible to either detect this somehow, or expose an option to ignore `updatedb.conf` entirely? As a sanity check, I will try to patch the CSI driver locally to temporarily skip the `updatedb.conf` mount for now to check if that is all that is required to get it working.

zwwhdls:
Hi @twelho , you can set an environment variable to control not to mount `updatedb.conf` into mount pod or any job.

twelho:
Sorry for the delay, but I can now confirm that JuiceFS works on Talos when commenting out the `updatedb.conf` mount, so that is the only blocker for immutable environments.

> Hi @twelho , you can set an environment variable to control not to mount `updatedb.conf` into mount pod or any job.

To my knowledge Kubernetes doesn't provide this kind of generic control through an environment variable, or are you suggesting that I should add an environment variable to the CSI driver to control this? I can refactor this PR do that if that's what you intend.

zwwhdls:
@twelho , I mean to add an ENV in juicefs csi node&controller to control if to mount `updatedb.conf` in mount pod when creating

twelho:
I've now added `JUICEFS_MODIFY_UPDATEDB` for controlling whether to mount `updatedb.cfg` or not. If not specified, it is enabled by default to keep the existing behavior. Right now I've only added it to the nodes and not the controllers, since only the nodes are responsible for spawning the mount pods. I've successfully tested this on a Talos Linux cluster by setting `JUICEFS_MODIFY_UPDATEDB=0` for the `juicefs-csi-node` DaemonSet (I'm using the Helm chart as a base).

twelho:
Now that I think about it, it might make sense to rename the variable to `JUICEFS_IMMUTABLE`, which could also cover any future aspects that JuiceFS might need to take into account in immutable environments if, for example, there should be another configuration file that needs to be modified on traditional, mutable systems, but which is restricted in immutable environments. Does this sound better? If we go with `JUICEFS_IMMUTABLE`, I will add it to both the CSI controllers and nodes.

zwwhdls:
@twelho sure, `JUICEFS_IMMUTABLE` is ok.

PR #729, opened 2023-08-29
enhance csi-doctor.sh collect

timfeirg:
* multi-pv support for the collect command
* add kubectl top information as well
* collect juicefs doctor results as well
* ignore non-jfs volumes

image

### DIFF
--- a/scripts/csi-doctor.sh
+++ b/scripts/csi-doctor.sh
@@ -272,6 +272,9 @@ collect_pv() {
     fi
   done
 
+  kubectl -n $juicefs_namespace cp ${BASH_SOURCE[0]} $mount_pod_name:/tmp
+  kubectl -n $juicefs_namespace exec $mount_pod_name -- /tmp/csi-doctor.sh doctor
+  kubectl -n $juicefs_namespace cp $mount_pod_name:/tmp/juicefs-doctor $diagnose_dir
 }
 
 collect_juicefs_csi_msg() {
@@ -331,6 +334,17 @@ collect() {
   rm -rf $diagnose_dir
 }
 
+doctor() {
+  mkdir -p /tmp/juicefs-doctor
+  cd /tmp/juicefs-doctor
+  mps=$(df -t fuse.juicefs --output=target | tail -n +2)
+  for mp in $mps
+  do
+    fname=$(juicefs doctor $mp | tail -n 1 | grep -oP '\S+.tar.gz')
+    mv $fname $(echo $mp | tr '/' '_').tar.gz
+  done
+}
+
 main() {
   if [[ $# -eq 0 ]]; then
     print_usage
@@ -349,6 +363,9 @@ main() {
         action=$1
         ;;
       collect|help)
+        action=$1
+        ;;
+      doctor|help)
         action=$1
         ;;
       get-mount|help)
@@ -377,6 +394,9 @@ main() {
       ;;
     collect)
       collect
+      ;;
+    doctor)
+      doctor
       ;;
     get-mount)
       get_mount_pod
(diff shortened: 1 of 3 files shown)


timfeirg:
I'll also add a `kubectl top` command to collect runtime resource usage for pods

scripts/csi-doctor.sh
```diff
@@ -257,21 +237,40 @@ mount_exec() {
...
-  $kbctl logs "$MOUNT_POD_NAME" -n $juicefs_namespace &>"$diagnose_dir/juicefs-$juicefs_namespace/mount-pod.log" 2>&1
-  $kbctl get po "$MOUNT_POD_NAME" -oyaml -n $juicefs_namespace &>"$diagnose_dir/juicefs-$juicefs_namespace/mount-pod.yaml" 2>&1
-  $kbctl describe po "$MOUNT_POD_NAME" -n $juicefs_namespace &>"$diagnose_dir/juicefs-$juicefs_namespace/mount-pod-describe.log" 2>&1
+  PVC_NAMES=$(${kbctl} -n ${namespace} get po ${app} -o jsonpath='{..persistentVolumeClaim.claimName}')
+  mkdir -p "$diagnose_dir/pv"
+  mkdir -p "$diagnose_dir/pvc"
+  for pvc_name in $PVC_NAMES
+  do
+
+    $kbctl get pvc "$pvc_name" -n $namespace -oyaml &>"$diagnose_dir/pvc/$pvc_name.yaml" 2>&1
```
zwwhdls, line 268: pvc may not be juicefs, if not, we should not collect it.

timfeirg: adopted in the latest commit

timfeirg:
I've added JuiceFS PV detection in the latest commit, non-jfs pvs will be ignored @zwwhdls

scripts/csi-doctor.sh
```diff
@@ -82,8 +63,21 @@ debug_app_pod() {
...
   for pvc_name in $PVC_NAMES
   do
+    pv_name=$(${kbctl} -n ${namespace} get pvc $pvc_name -o jsonpath='{.spec.volumeName}' 2>/dev/null)
+
+    if [ "$pv_name" == "" ]; then
+      echo "Ignore $pvc_name due to empty volumeName"
+      continue
+    fi
+
+    provisioner=$(${kbctl} get pv ${pv_name} -o 'go-template={{index .metadata.annotations "pv.kubernetes.io/provisioned-by"}}' 2>/dev/null)
```
zwwhdls: If pv is created by users(static provision), its value is `pv.kubernetes.io/bound-by-controller: "yes"`. 

I recommend to use `pv.csi.driver == "csi.juicefs.com"` instead of annotation.

scripts/csi-doctor.sh
```diff
@@ -257,22 +251,55 @@ mount_exec() {
...
+    $kbctl describe pvc "$pvc_name" -n $namespace &>"$diagnose_dir/pvc/$pvc_name-describe.log" 2>&1
+
+    pv_name=$(${kbctl} -n ${namespace} get pvc $pvc_name -o jsonpath='{.spec.volumeName}')
+
+    if [ "$pv_name" == "" ]; then
+      echo "Ignore $pvc_name due to empty volumeName"
+      continue
+    fi
 
+    provisioner=$(${kbctl} get pv ${pv_name} -o 'go-template={{index .metadata.annotations "pv.kubernetes.io/provisioned-by"}}')
```
zwwhdls: ditto

zwwhdls (changes requested):

scripts/csi-doctor.sh
timfeirg: good point, fixed

zwwhdls (approved):

Issue #738, opened 2023-09-06
[BUG] The chunk files remain in the object storage even if the PVC is deleted


kyungwan-nam:
**What happened**:

The chunk files remain in the object storage even if the PVC is deleted

**What you expected to happen**:

``reclaimPolicy`` is set as ``Delete`` in the storageclass.
I expected to clean the object storage as well as the metadata.
but, the chunk files remain in the object storage.

**How to reproduce it (as minimally and precisely as possible)**:

**Anything else we need to know?**

**Environment**:
- JuiceFS CSI Driver version (which image tag did your CSI Driver use):
- Kubernetes version (e.g. `kubectl version`):
- Object storage (cloud provider and region):
- Metadata engine info (version, cloud provider managed or self maintained):
- Network connectivity (JuiceFS to metadata engine, JuiceFS to object storage):
- Others:

Hexilee:
https://github.com/juicedata/juicefs-csi-driver/pull/739#issuecomment-1711019449

> I do not recommend to execute `gc` automatically because of some reasons:
> 
> 1. `gc` will scan the whole file system, including metadata and object storage, which may be very slow.
> 2. Even if you executed the `gc` command, the chunk files may still exist in the object storage because JuiceFS enables the trash directory by default.
> 
> So, I recommend you to clean up the trash directory as necessary and run `gc` command mannelly.

Hexilee:
Resolved in https://github.com/juicedata/juicefs-csi-driver/pull/739

PR #739, opened 2023-09-06
fix: The chunk files remain in the object storage even if the PVC is …

kyungwan-nam:
…deleted

https://github.com/juicedata/juicefs-csi-driver/issues/738

The job for volume deletion will be terminated immediately after running ``juicefs rmr``.

I added ``juicefs gc --delete``to clean the data in the object storage.

### DIFF
--- a/pkg/juicefs/mount/builder/job.go
+++ b/pkg/juicefs/mount/builder/job.go
@@ -130,12 +130,15 @@ func (r *Builder) getCreateVolumeCmd() string {
 func (r *Builder) getDeleteVolumeCmd() string {
 	cmd := r.getJobCommand()
 	var jfsPath string
+	var gcArgs string
 	if r.jfsSetting.IsCe {
 		jfsPath = config.CeCliPath
+		gcArgs = "${metaurl}"
 	} else {
 		jfsPath = config.CliPath
+		gcArgs = "${VOL_NAME}"
 	}
-	return fmt.Sprintf("%s && if [ -d /mnt/jfs/%s ]; then %s rmr /mnt/jfs/%s && %s gc ${metaurl} --delete; fi;", cmd, r.jfsSetting.SubPath, jfsPath, r.jfsSetting.SubPath, jfsPath)
+	return fmt.Sprintf("%s && if [ -d /mnt/jfs/%s ]; then %s rmr /mnt/jfs/%s && %s gc %s --delete; fi;", cmd, r.jfsSetting.SubPath, jfsPath, r.jfsSetting.SubPath, jfsPath, gcArgs)
 }
 
 func (r *Builder) getJobCommand() string {
--- a/pkg/juicefs/mount/builder/job.go
+++ b/pkg/juicefs/mount/builder/job.go
@@ -135,7 +135,7 @@ func (r *Builder) getDeleteVolumeCmd() string {
 	} else {
 		jfsPath = config.CliPath
 	}
-	return fmt.Sprintf("%s && if [ -d /mnt/jfs/%s ]; then %s rmr /mnt/jfs/%s; fi;", cmd, r.jfsSetting.SubPath, jfsPath, r.jfsSetting.SubPath)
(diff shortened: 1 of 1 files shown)


pkg/juicefs/mount/builder/job.go
```diff
@@ -135,7 +135,7 @@ func (r *Builder) getDeleteVolumeCmd() string {
 	} else {
 		jfsPath = config.CliPath
 	}
-	return fmt.Sprintf("%s && if [ -d /mnt/jfs/%s ]; then %s rmr /mnt/jfs/%s; fi;", cmd, r.jfsSetting.SubPath, jfsPath, r.jfsSetting.SubPath)
+	return fmt.Sprintf("%s && if [ -d /mnt/jfs/%s ]; then %s rmr /mnt/jfs/%s && %s gc ${metaurl} --delete; fi;", cmd, r.jfsSetting.SubPath, jfsPath, r.jfsSetting.SubPath, jfsPath)
```
zwwhdls: `gc` is a little different between [ce](https://juicefs.com/docs/zh/community/command_reference#gc) (`juicefs gc $METAURL --delete`) and [ee](https://juicefs.com/docs/zh/cloud/reference/command_reference#gc) (`juicefs gc $VOL_NAME --delete`).

zwwhdls (changes requested):

pkg/juicefs/mount/builder/job.go
kyungwan-nam: I fixed it.
I can't test it under ``ee``, but it works well on ``ce``.

pkg/juicefs/mount/builder/job.go
```diff
@@ -130,12 +130,15 @@ func (r *Builder) getCreateVolumeCmd() string {
 func (r *Builder) getDeleteVolumeCmd() string {
 	cmd := r.getJobCommand()
 	var jfsPath string
+	var gcArgs string
 	if r.jfsSetting.IsCe {
 		jfsPath = config.CeCliPath
+		gcArgs = "${metaurl}"
 	} else {
 		jfsPath = config.CliPath
+		gcArgs = "${VOL_NAME}"
```
zwwhdls: You can use `r.jfsSetting.Name`

Hexilee:
I do not recommend to execute `gc` automatically because of some reasons:

1. `gc` will scan the whole file system, including metadata and object storage, which may be very slow.
2. Even if you executed the `gc` command, the chunk files may still exist in the object storage because JuiceFS enables the trash directory by default.

So, I recommend you to clean up the trash directory as necessary and run `gc` command mannelly.

kyungwan-nam:
@Hexilee 

Thanks for your comment! It makes sense.
I think ``trash`` is not useful on [dynamic-provisioning](https://juicefs.com/docs/csi/guide/pv#dynamic-provisioning)
Users will expect to clean their data when deleting PVC which is ``Delete`` reclaimPolicy. Also, It doesn't need to be recovered. so, the trash is not enabled in our cluster.
It would be good if it is configurable not default. please let me know your thoughts. thanks!

Hexilee:
> @Hexilee
> 
> Thanks for your comment! It makes sense. I think `trash` is not useful on [dynamic-provisioning](https://juicefs.com/docs/csi/guide/pv#dynamic-provisioning) Users will expect to clean their data when deleting PVC which is `Delete` reclaimPolicy. Also, It doesn't need to be recovered. so, the trash is not enabled in our cluster. It would be good if it is configurable not default. please let me know your thoughts. thanks!

Well, if you disabled the trash directory by `juicefs config <metaurl> --trash-days 0`, the chunk objects are also deleted when files are deleted. However, because the objects are deleted asynchronously, you may wait for some time to see the results. 

In this case, you should execute `gc` command only when there are some leaked objects.

kyungwan-nam:
@Hexilee 

> However, because the objects are deleted asynchronously, you may wait for some time to see the results.

Even if I wait, it may not be deleted. because the juices client does not exist after the PVC is deleted.

Hexilee:
> deleted

Other juicefs clients will continue to do it. You can execute `gc` manually If there is no other active PVC in a long time.

kyungwan-nam:
@Hexilee @zwwhdls 
Thank you for letting me know.
I understood this causes bad effects as your comment.
I will try another way to solve it. thanks!

Hexilee:
@kyungwan-nam Thanks and look forward to your contribution!

PR #746, opened 2023-09-11
revert #741: bind fuse connections into mount pod

### LINKED ISSUE #744: [BUG] mount failed when `/sys/fs/fuse` not exist in host
**What happened**:
```
Warning  FailedMount  1s (x7 over 33s)  kubelet  MountVolume.SetUp failed for volume "fuse-connections" : mkdir /sys/fs/fuse: operation not permitted
```

**Environment**:
- JuiceFS CSI Driver version (which image tag did your CSI Driver use): v0.22.0
- Others:

Hexilee:
close #744

The `/sys/fs/fuse/connections` may not exist because of a [kernel bug](https://github.com/systemd/systemd/issues/4574)

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -51,8 +51,6 @@ var (
 	JFSMountPriorityName     = "system-node-critical"
 	JFSMountPreemptionPolicy = ""
 
-	FuseConnectionPath = "/sys/fs/fuse/connections"
-
 	TmpPodMountBase = "/tmp"
 	PodMountBase    = "/jfs"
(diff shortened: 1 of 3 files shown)


Hexilee:
Refer to https://github.com/ceph/ceph/pull/34839, execute `modprobe fuse` manually may fix this issue. However, the fuse module in a container may conflict with the kernel.

Binding the whole `/sys/fs` may work, but the `/sys/fs` already exists in `jfs-mount` container, overriding them may cause unexpected conditions.

PR #753, opened 2023-09-18
support vci serverless

zwwhdls:
1. when in serverless environment, do not mount hostPath except mount point. Set `juicefs.com/enable-serverless-injection: "true"` in namespace to use.
2. support VCI serverless. Set `juicefs.com/enable-serverless-injection: "true"` in namespace and `vke.volcengine.com/burst-to-vci: enfore` in pod to use.
    -  do not mount hostPath
    -  container is not privileged
    -  no propagationBidirectional mountPropagation
    -  no initContainer
    -  with env JFS_NO_UMOUNT=1
    -  annotations for VCI


Hexilee (approved):
LGTM

PR #755, opened 2023-09-20
mount pod: use node selector instead of node name

### LINKED ISSUE #754: mount pod: use node selector instead of node name

Hexilee:
close #754


zwwhdls:
Please fix when pod error because of lack of resources.

PR #785, opened 2023-11-01
mount pod: inherit affinity, schedulerName and tolerations from app pod

Hexilee:


squat:
Does the "app pod" mean the pod that is requesting the PVC?

Issue #773, opened 2023-10-26
[BUG] invalid format options


jianghushinian:
**What happened**:

When I use temporary access credentials for Huawei obs, I get the following error:

`invalid format options: xxx`

**What you expected to happen**:

no errors.

**How to reproduce it (as minimally and precisely as possible)**:

Secret YAML example:

```yaml
apiVersion: v1
kind: Secret
metadata:
  name: juicefs-secret
  namespace: kube-system
type: Opaque
stringData:
  name: juicefs-test
  metaurl: redisurl
  storage: obs
  bucket: obs-bucket
  access-key: ak
  secret-key: sk
  format-options: session-token=xxxxxxx=
```

Because the session-token of Huawei obs ends with `=`, so `JfsSetting.ParseFormatOptions` method verification will make an error, code show as below, `len(pair)` will be 3:

https://github.com/juicedata/juicefs-csi-driver/blob/80bc0f6f5c34c492e50f9e8f141cff4d1e3eb672/pkg/config/setting.go#L370-L390

**Anything else we need to know?**

**Environment**:
- JuiceFS CSI Driver version (which image tag did your CSI Driver use):
- Kubernetes version (e.g. `kubectl version`):
- Object storage (cloud provider and region):
- Metadata engine info (version, cloud provider managed or self maintained):
- Network connectivity (JuiceFS to metadata engine, JuiceFS to object storage):
- Others:

showjason:
Hi @jianghushinian, maybe you can try the following secret yaml(but I'm not sure whether it will be working, that's my guess):)
```
apiVersion: v1
kind: Secret
metadata:
  name: juicefs-secret
  namespace: kube-system
type: Opaque
stringData:
  name: juicefs-test
  metaurl: redisurl
  storage: obs
  bucket: obs-bucket
  access-key: ak
  secret-key: sk
  session-token: xxxxxxx=
```

Hexilee:
closed by #786

PR #786, opened 2023-11-01
to fix the format option issue

showjason:
#773 
I see the subcommand `format` supports multiple secret key-like options whose value will contain `=` most likely, therefore I try to use `SplitN` to split the options to fixed length 2, so that the `=` in the value can be ignored. If it's not reasonable, please let me know


zwwhdls:
Hi @showjason , please fix the unit test. thx!

showjason:
> Hi @showjason , please fix the unit test. thx!

Ah...sorry, I missed some particular cases, let me fix.

zwwhdls:
@Hexilee PTAL

Hexilee (approved):

PR #798, opened 2023-11-15
enhance quota: runs in background when block; no fail

Hexilee:


pkg/juicefs/juicefs.go
```diff
@@ -806,13 +806,33 @@ func (j *juicefs) SetQuota(ctx context.Context, secrets map[string]string, jfsSe
...
+			// ce cli will block until quota is set
+			res, err = j.Exec.CommandContext(context.Background(), config.CeCliPath, args...).CombinedOutput()
+			close(done)
+		}()
+		select {
+		case <-cmdCtx.Done():
+			go func() {
+				klog.Warningf("quota set timeout, runs in background")
+				<-done
+				if err := wrapSetQuotaErr(string(res), err); err != nil {
```
zwwhdls: There is no need to start a new goroutine to collect logs. You can print logs directly in the `set qupta` goroutine.

Hexilee: code is polished

zwwhdls (approved):

PR #801, opened 2023-11-17
fix: get pod by pv volumeHandle

zwwhdls:
1. get all pvc when multiple pvs has the same volumeHandle when list app pods
2. requeue mount pod periodically to avoid leaking in both kubelet polling and apiserver watching


pkg/controller/pod_controller.go
```diff
@@ -84,16 +84,18 @@ func (m *PodController) Reconcile(ctx context.Context, request reconcile.Request
 			continue
 		}
 		if uniqueId != "" && (p.Spec.CSI.VolumeHandle == uniqueId || p.Spec.StorageClassName == uniqueId) {
-			pv = &p
-			break
+			relatedPVs = append(relatedPVs, &p)
```
Hexilee, line 88: We should throw a warning when the same `VolumeHandle` is used by multiple PVs.

Hexilee (approved):

PR #839, opened 2023-12-25
recover: umount target when count > 5

zwwhdls:


pkg/controller/pod_driver.go
```diff
@@ -569,12 +574,48 @@ func (p *PodDriver) recoverTarget(podName, sourcePath string, ti *targetItem, mi
...
+	klog.V(5).Infof("umount target %d times", count)
 	for i := 0; i < count; i++ {
 		// ignore error
 		p.Unmount(target)
 	}
 }
 
+// umountTargetUntilRemain umount target path with remaining count
+func (p *PodDriver) umountTargetUntilRemain(basemi *mountItem, target string, remainCount int) {
+	for {
```
Hexilee, line 595: Maybe we need a timeout for this loop

Hexilee (approved):

PR #843, opened 2023-12-28
fix advisory: shell script injection 

Hexilee:


pkg/util/security/escape_test.go
```diff
@@ -0,0 +1,41 @@
...
+
+func TestEscapeBashStr(t *testing.T) {
+	cases := [][]string{
+		{"abc", "abc"},
+		{"test-volume", "test-volume"},
+		{"http://minio.kube-system:9000/minio/dynamic-ce", "http://minio.kube-system:9000/minio/dynamic-ce"},
+		{"$(cat /proc/self/status | grep CapEff > /test.txt)", "$'$(cat /proc/self/status | grep CapEff > /test.txt)'"},
+		{"hel`cat /proc/self/status`lo", "$'hel`cat /proc/self/status`lo'"},
+		{"'h'el`cat /proc/self/status`lo", "$'\\'h\\'el`cat /proc/self/status`lo'"},
+		{"\\'h\\'el`cat /proc/self/status`lo", "$'\\'h\\'el`cat /proc/self/status`lo'"},
```
davies: We could implement `unescape` for test.

Hexilee: We can straightly `unescape` it by `bash -c "echo %s"`

davies: Sounds good

Hexilee: Done

PR #845, opened 2024-01-07
cache auth results and use as client conf

timfeirg:
goal: CSI Driver will continue to work even if connection towards web console is disrupted

implementation:

* CSI Controller will watch all secrets, if a JuiceFS volume credentials is created/updated, it'll run `juicefs auth` within the controller container, and put the client conf file into the `init_config` field inside the Kubernetes secret
* if `init_config` is detected, an extra volume mount will be created inside the mount pod, i.e. `myjfs.conf -> /etc/juicefs/myjfs.conf`
* when mount pod starts, client conf will be copied into `/root/.juicefs`, thus even if auth fails, a valid client conf will be used to ensure continuous service

caveats:

* controller must be up and running for this feature to work, even if dynamic provisioning isn't used
* not yet support the situation where multiple credential secrets use a same JuiceFS Volume, client conf files will overwrite each other

simple tests are run to verify that this actually work:

```
k logs --all-containers -f juicefs-chaos-k8s-001-chaos-ee-test-kybcwp
2024/01/07 07:03:47.875750 juicefs[8] <WARNING>: update config: POST https://juicefs.com/volume/chaos-ee-test/mount: Post "https://juicefs.com/volume/chaos-ee-test/mount": context deadline exceeded (Client.Timeout exceeded while awaiting headers), use existed one [auth.go:616]
2024/01/07 07:03:49.495372 juicefs[54] <INFO>: JuiceFS version 5.0.6 (2023-12-25 04eaeb2) [mount.go:629]
2024/01/07 07:03:49.534196 juicefs[54] <INFO>: Cache: /var/jfsCache/chaos-ee-test capacity: 102400 MB [disk_cache.go:1074]
2024/01/07 07:03:49.534902 juicefs[54] <INFO>: Mounting volume chaos-ee-test at /jfs/chaos-ee-test-kybcwp ... [mount_unix.go:651]
OK, chaos-ee-test is ready at /jfs/chaos-ee-test-kybcwp.
2024/01/07 07:03:54.228262 juicefs[26] <INFO>: watching /jfs/chaos-ee-test-kybcwp, pid 54 [mount_unix.go:112]
```

PR #847, opened 2024-01-10
introduce feature flag --cache-client-conf

timfeirg:
* use `--cache-client-conf` as feature flag for https://github.com/juicedata/juicefs-csi-driver/pull/845
* use initconfig instead of init_config in both secrets (the secret created by user, and the one created by csi driver)
* do not modify secret once initconfig has been set

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -35,6 +35,7 @@ var (
 	process     bool
 
 	provisioner bool
+	cacheConf   bool
 	webhook     bool
 	certDir     string
 	webhookPort int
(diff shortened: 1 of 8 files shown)


cmd/main.go
```diff
@@ -67,6 +68,7 @@ func main() {
 
 	// controller flags
 	cmd.Flags().BoolVar(&provisioner, "provisioner", false, "Enable provisioner in controller. default false.")
+	cmd.Flags().BoolVar(&cacheConf, "cache-client-conf", false, "Cache client config file. default false.")
```
davies, line 71: how to use this one?

We should add a test case for it

timfeirg:
close in favor of https://github.com/juicedata/juicefs-csi-driver/pull/850

fork pr doesn't run e2e tests

PR #850, opened 2024-01-12
introduce feature flag --cache-client-conf

timfeirg:
* use --cache-client-conf as feature flag for #845
* use initconfig instead of init_config in both secrets (the secret created by user, and the one created by csi driver)
* do not modify secret once initconfig has been set

### DIFF
--- a/deploy/kubernetes/csi-ci/pod/statefulset.yaml
+++ b/deploy/kubernetes/csi-ci/pod/statefulset.yaml
@@ -13,4 +13,5 @@ spec:
             - --logtostderr
             - --nodeid=$(NODE_NAME)
             - --leader-election
+            - --cache-client-conf
             - --v=6
--- a/deploy/kubernetes/csi-ci/pod/statefulset.yaml
+++ b/deploy/kubernetes/csi-ci/pod/statefulset.yaml
@@ -13,5 +13,4 @@ spec:
             - --logtostderr
             - --nodeid=$(NODE_NAME)
             - --leader-election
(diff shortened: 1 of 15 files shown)


deploy/kubernetes/csi-ci/pod/statefulset.yaml
```diff
@@ -13,4 +13,5 @@ spec:
             - --logtostderr
             - --nodeid=$(NODE_NAME)
             - --leader-election
+            - --cache-client-conf
```
timfeirg, line 16: I'll add more tests on this function

timfeirg:
考虑到目标是稳定以后作为默认开启的 feature，因此目前只是在 share pod 模式“偷偷”加上了这个 feature flag，让端对端测试可以“顺便”测到这个功能。

之后经过大量验证、稳定运行后，将默认值改为 true，测试不需要进行变动。 @davies

pkg/controller/secret_controller.go
```diff
@@ -124,12 +126,18 @@ func (m *SecretController) SetupWithManager(mgr ctrl.Manager) error {
 				klog.V(6).Info("secret.onUpdateFunc Skip due to resourceVersion not changed")
 				return false
 			}
+
+			_, existsNew := secretNew.Data["initconfig"]
```
Hexilee: I think we have better checking `initconfig` of `secretOld` instead of `secretNew`, as our goal is "generate initconfig only once"

Hexilee (approved):

PR #855, opened 2024-01-23
fix: cut the pod key if the length is greater than `maxPodLabelKeyLen…

### LINKED ISSUE #854: [BUG] long namespace is not supported
When namespace is long, the generated pod name may exceed limit, we should only use the part of it when it's too long.

zxh326:
fix: #854

### DIFF
--- a/pkg/util/util.go
+++ b/pkg/util/util.go
@@ -42,6 +42,7 @@ import (
 
 const (
 	maxListTries                         = 3
+	maxPodLabelKeyLength                 = 63
 	expectedAtLeastNumFieldsPerMountInfo = 10
 	procMountInfoPath                    = "/proc/self/mountinfo"
 )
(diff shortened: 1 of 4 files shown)


davies (approved):

pkg/util/util.go
```diff
@@ -448,3 +449,11 @@ func ImageResol(image string) (hasCE, hasEE bool) {
 	}
 	return true, true
 }
+
+// CutPodLabelKey removes the suffix of the key if the length is greater than `maxPodLabelKeyLength`
+func CutPodLabelKey(key string) string {
+	if len(key) > maxPodLabelKeyLength {
+		return key[:maxPodLabelKeyLength]
```
davies, line 456: Will it be better to keep the last part?

davies: to reduce conflicts

pkg/util/util_test.go
```diff
@@ -625,3 +625,22 @@ func TestImageResol(t *testing.T) {
...
 }
+
+func TestCutPodLabelKey(t *testing.T) {
+	tests := []struct {
+		name string
+		key  string
+		want string
+	}{
+		{"normal length of key label", "label_key_1", "label_key_1"},
+		{"longest length of key label ", "label_key_test-xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx", "label_key_test-xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx"},
```
timfeirg, line 636: consider using random string to ensure the back part is being preserved

PR #867, opened 2024-02-02
use set +e in github workflow script

timfeirg:
in a [recent build](https://github.com/juicedata/juicefs-csi-driver/actions/runs/7749403577/job/21133881274), FUSE_IMAGE_EXIST wasn't correctly evaluated.

the reason seems to be `shell: /usr/bin/bash -e {0}` making the script running with `-e`, which abort at any non-zero code.

in this pr, `set +e` to let script continue, and manually exit when needed


zxh326 (approved):

chenjie4255 (approved):
确认中间过程的重要错误都已处理即可。不过我还是偏向加的保守一点，例如哪里出错了再在哪里加

PR #870, opened 2024-02-06
feat: overwrite mount pod resources with PVC annotations

zxh326:


timfeirg:
since this pr introduces annotation based resource declaration, there's no reason to continue using [`volumeAttributes`](https://juicefs.com/docs/csi/guide/resource-optimization/#mount-pod-resources), correct?

if that's the case, we should recommend annotation as the first option for resources declaration

zxh326:
> since this pr introduces annotation based resource declaration, there's no reason to continue using [`volumeAttributes`](https://juicefs.com/docs/csi/guide/resource-optimization/#mount-pod-resources), correct?
> 
> if that's the case, we should recommend annotation as the first option for resources declaration

Now we have two ways to customize the mount pod resource.

1. Static PV, volumeAttributes

2. Using parameters of storageClass

these two methods can be modified later.

We can consider not abandoning these two ways for now, and suggest customizing the mount pod resource using pvc annotations in the documentation.

timfeirg:
the old ways will certainly be preserved, I'll remember to create a docs pr after this is released

pkg/juicefs/juicefs.go
```diff
@@ -339,6 +343,38 @@ func (j *juicefs) JfsMount(ctx context.Context, volumeID string, target string,
...
+		klog.Errorf("Get pvc %s/%s error: %v", pv.Spec.ClaimRef.Namespace, pv.Spec.ClaimRef.Name, err)
+		return err
+	}
+
+	cpuLimit := pvc.Annotations[config.MountPodCpuLimitKey]
+	memoryLimit := pvc.Annotations[config.MountPodMemLimitKey]
+	cpuRequest := pvc.Annotations[config.MountPodCpuRequestKey]
+	memoryRequest := pvc.Annotations[config.MountPodMemRequestKey]
+
+	if cpuLimit != "" {
```
davies: We can use a for-loop here

davies (approved):

pkg/controller/pod_driver.go
```diff
@@ -375,6 +375,9 @@ func (p *PodDriver) podDeletedHandler(ctx context.Context, pod *corev1.Pod) erro
 				}
 				controllerutil.AddFinalizer(newPod, config.Finalizer)
 				klog.Infof("Need to create pod %s %s", pod.Name, pod.Namespace)
+				if err := p.OverwirteMountPodResourcesWithPVC(ctx, newPod); err != nil {
+					klog.Errorf("overwrite mount pod resources wiht pv error %v", err)
```
timfeirg: typo

pkg/juicefs/juicefs.go
```diff
@@ -318,6 +318,10 @@ func (j *juicefs) JfsMount(ctx context.Context, volumeID string, target string,
 	if err := j.validTarget(target); err != nil {
 		return nil, err
 	}
+
+	if err := j.overwriteVolCtxWithPVCAnnotations(ctx, volumeID, volCtx); err != nil {
+		klog.Warningf("Overwrite volCtx with PVC annotations error: %v", err)
```
timfeirg: maybe use a log level of error since this will definitely go against user's expectation?

timfeirg (approved):

pkg/controller/pod_driver.go
```diff
@@ -376,7 +376,7 @@ func (p *PodDriver) podDeletedHandler(ctx context.Context, pod *corev1.Pod) erro
 				controllerutil.AddFinalizer(newPod, config.Finalizer)
 				klog.Infof("Need to create pod %s %s", pod.Name, pod.Namespace)
 				if err := p.OverwirteMountPodResourcesWithPVC(ctx, newPod); err != nil {
-					klog.Errorf("overwrite mount pod resources wiht pv error %v", err)
+					klog.Errorf("Overwrite mount pod resources wiht pvc error %v", err)
```
timfeirg: sorry I mean

```suggestion
					klog.Errorf("Overwrite mount pod resources with pvc error %v", err)
```

timfeirg:
merge and I'll handle the docs update in a separate PR @zxh326

PR #875, opened 2024-02-20
fix: create the secret before start the mountpod

zxh326:


timfeirg:
tests are failing due to cache space, maybe rebase origin/master and then retry? @zxh326

timfeirg (approved):

PR #887, opened 2024-03-04
fix: use lazy umount

zxh326:
Signed-off-by: Xuhui zhang <xuhui@juicedata.io>

### DIFF
--- a/pkg/controller/pod_driver.go
+++ b/pkg/controller/pod_driver.go
@@ -82,7 +82,7 @@ func (p *PodDriver) SetMountInfo(mit mountInfoTable) {
 
 func (p *PodDriver) Run(ctx context.Context, current *corev1.Pod) error {
 	podStatus := getPodStatus(current)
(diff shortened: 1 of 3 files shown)


pkg/controller/pod_driver.go
```diff
@@ -81,13 +81,14 @@ func (p *PodDriver) SetMountInfo(mit mountInfoTable) {
 }
 
 func (p *PodDriver) Run(ctx context.Context, current *corev1.Pod) error {
+	podStatus := getPodStatus(current)
+	klog.V(6).Infof("[PodDriver] start handle pod %s/%s, status: %s", current.Name, current.Namespace, podStatus)
```
timfeirg: stick to print output convention?

```suggestion
	klog.V(6).Infof("[PodDriver] start handle pod %s/%s, status: %s", current.Namespace, current.Name,  podStatus)
```

timfeirg (approved):

PR #889, opened 2024-03-05
Fix the sidecar mutate when using multiple PVCs in a single pod

zxh326:


pkg/juicefs/mount/builder/vci-serverless.go
```diff
@@ -64,13 +71,27 @@ func NewVCIBuilder(setting *config.JfsSetting, capacity int64, app corev1.Pod, p
...
+	propagations = append(propagations, VCIPropagationStruct{
+		Container: mountContainerName,
+		MountPath: r.jfsSetting.MountPath,
+	})
+	if v, ok := r.app.Annotations[VCIPropagation]; ok {
+		var vciPropagations []VCIPropagationStruct
+		if err := json.Unmarshal([]byte(v), &vciPropagations); err == nil {
+			propagations = append(propagations, vciPropagations...)
+		} else {
+			klog.Errorf("failed to unmarshal VCIPropagation annotation: %v", err)
```
timfeirg, line 90: any change we print the original annotations string in this log?

```suggestion
			klog.Errorf("failed to unmarshal VCIPropagation annotation %s %v", v, err)
```

pkg/juicefs/mount/builder/vci-serverless.go
```diff
@@ -64,13 +71,27 @@ func NewVCIBuilder(setting *config.JfsSetting, capacity int64, app corev1.Pod, p
 func (r *VCIBuilder) NewMountSidecar() *corev1.Pod {
```
timfeirg, line 71: if not too much trouble, could you apply the changes to all `pkg/juicefs/mount/builder/*serverless.go` codes? I imagine the code will be all similar?

this way, if tests pass for `pkg/juicefs/mount/builder/serverless.go`, we're quite confident that it's going to work for other platforms as well?

zxh326: only vci requires this annotation

pkg/juicefs/mount/builder/vci-serverless.go
zxh326: emmm...I believe there is no need to print

pkg/juicefs/mount/builder/vci-serverless.go
zxh326: *serverless.go have a lot of similar code, which can be optimized later.

timfeirg (approved):
测通以后自行合并

PR #890, opened 2024-03-06
Add retry logic for reconciler of polling kubelet & ignore `No mount point specified` error  in preStop

zxh326:
- Add retry logic for reconciler of polling kubelet
- fix for-range case in api-server reconciler
- ignore errors in preStop hook

### DIFF
--- a/pkg/controller/pod_driver.go
+++ b/pkg/controller/pod_driver.go
@@ -130,12 +130,6 @@ func (p *PodDriver) checkAnnotations(ctx context.Context, pod *corev1.Pod) error
 	lock.Lock()
 	defer lock.Unlock()
 
-	// Add a 10s delay to avoid misjudgment of deletion due to cache reasons when the pod is just created
-	if time.Now().Before(pod.GetCreationTimestamp().Time.Add(10 * time.Second)) {
-		klog.V(5).Infof("[PodDriver] pod %s is just created, skip check annotations", pod.Name)
-		return nil
-	}
-
 	delAnnotations := []string{}
 	var existTargets int
 	for k, target := range pod.Annotations {
--- a/pkg/controller/pod_driver.go
+++ b/pkg/controller/pod_driver.go
@@ -135,8 +135,10 @@ func (p *PodDriver) checkAnnotations(ctx context.Context, pod *corev1.Pod) error
 	for k, target := range pod.Annotations {
 		if k == util.GetReferenceKey(target) {
 			targetUid := getPodUid(target)
+			// only it is not in pod lists can be seen as deleted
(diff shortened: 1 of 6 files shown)


pkg/juicefs/mount/builder/common.go
```diff
@@ -99,7 +99,7 @@ func (r *BaseBuilder) genCommonJuicePod(cnGen func() corev1.Container) *corev1.P
 	pod.Spec.Containers[0].Lifecycle = &corev1.Lifecycle{
 		PreStop: &corev1.Handler{
 			Exec: &corev1.ExecAction{Command: []string{"sh", "-c", fmt.Sprintf(
-				"umount %s -l && rmdir %s", r.jfsSetting.MountPath, r.jfsSetting.MountPath)}},
+				"umount %s -l || [ $? -eq 32 ] && rmdir %s || [ $? -eq 1 ]", r.jfsSetting.MountPath, r.jfsSetting.MountPath)}},
```
timfeirg: can we 

```
sh -c +e "umount -l xxx; rmdir xxx; exit 0;"
```

to essentially ignore all errors? since prestop hook failures doesn't affect outcome, and we typically don't investigate individual hook errors

timfeirg (approved):

pkg/controller/pod_driver.go
```diff
@@ -135,8 +135,10 @@ func (p *PodDriver) checkAnnotations(ctx context.Context, pod *corev1.Pod) error
 	for k, target := range pod.Annotations {
 		if k == util.GetReferenceKey(target) {
 			targetUid := getPodUid(target)
+			// Only it is not in pod lists can be seen as deleted
+			// Add a 10s delay to avoid misjudgment of deletion due to cache reasons when the pod is just created.
```
timfeirg: use consistent ending dot 😅

```suggestion
			// Add a 10s delay to avoid misjudgment of deletion due to cache reasons when the pod is just created
```

timfeirg (approved):

pkg/controller/pod_driver.go
```diff
@@ -130,13 +130,20 @@ func (p *PodDriver) checkAnnotations(ctx context.Context, pod *corev1.Pod) error
 	lock.Lock()
 	defer lock.Unlock()
 
+	// Add a 10s delay to avoid misjudgment of deletion due to cache reasons when the pod is just created
```
timfeirg: we're not sure of the root cause yet, AFAIK if no `ResourceVersion="0"` is set, the apiserver always query etcd to get the latest state. 

```suggestion
	// Add a 10s delay to avoid misjudgment of deletion due to rare ApiServer info mismatch when the pod is just created
```

PR #896, opened 2024-03-13
Upgrade default ee mount version to the latest

zxh326:
The current version is `5.0.10`.


timfeirg (approved):
FYI @davies @chnliyong

PR #905, opened 2024-03-21
feat: mount pod probe

zxh326:


timfeirg:
I suggest only implement readinessProbe for now, we have no reliable ways to determine true liveness (wether a restart can resolve the problem, or no effect at all).

also, mount is relatively fast so no need for startupProbe

PR #908, opened 2024-03-26
Allow set mountpod to zero resources

zxh326:
after https://github.com/juicedata/juicefs-csi-driver/pull/771 when resource is set smaller than 0, will not set resource request/limit

maybe some people want to set it to 0 resources. we should allow this

### DIFF
--- a/pkg/config/setting.go
+++ b/pkg/config/setting.go
@@ -519,7 +519,7 @@ func ParsePodResources(cpuLimit, memoryLimit, cpuRequest, memoryRequest string) 
 			return corev1.ResourceRequirements{}, err
 		}
 		q := podLimit[corev1.ResourceCPU]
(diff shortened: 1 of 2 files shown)


timfeirg:
originally, https://github.com/juicedata/juicefs-csi-driver/pull/771 wishes to let users "omit" the resource limits fields, so that a mount pod can use as many resource as it needs. this behavior makes sense and should be preserved.

however, setting requests to 0 is ambiguous, both to users & kubernetes itself (AFAIK some kubernetes distribution even disallow zero requests), but some use cases do need to use 0, so I think we should:

* when limits == 0, remove the limits field
* when requests == 0, pass through as-is, but this require kubernetes to interpret 0 as an actual 0, not node resource ceiling instead
* if we verified standard kubernetes handles 0 as node resource ceiling, I don't think there's anything we can do here

@zxh326

PR #909, opened 2024-03-27
feat: support for extra pre-stop hook command

zxh326:
doc updated later

### DIFF
--- a/pkg/config/setting.go
+++ b/pkg/config/setting.go
@@ -56,14 +56,15 @@ type JfsSetting struct {
 	ClientConfPath     string               `json:"-"`
 
 	// put in secret
-	SecretKey     string            `json:"secret-key,omitempty"`
-	SecretKey2    string            `json:"secret-key2,omitempty"`
-	Token         string            `json:"token,omitempty"`
-	Passphrase    string            `json:"passphrase,omitempty"`
-	Envs          map[string]string `json:"envs_map,omitempty"`
-	EncryptRsaKey string            `json:"encrypt_rsa_key,omitempty"`
-	InitConfig    string            `json:"initconfig,omitempty"`
-	Configs       map[string]string `json:"configs_map,omitempty"`
+	SecretKey           string            `json:"secret-key,omitempty"`
+	SecretKey2          string            `json:"secret-key2,omitempty"`
+	Token               string            `json:"token,omitempty"`
+	Passphrase          string            `json:"passphrase,omitempty"`
+	Envs                map[string]string `json:"envs_map,omitempty"`
+	EncryptRsaKey       string            `json:"encrypt_rsa_key,omitempty"`
+	InitConfig          string            `json:"initconfig,omitempty"`
+	Configs             map[string]string `json:"configs_map,omitempty"`
+	ExtraPreStopHookCmd string            `json:"-,omitempty"`
 
 	// put in volCtx
 	MountPodLabels      map[string]string `json:"mount_pod_labels"`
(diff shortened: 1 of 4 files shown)


pkg/config/setting.go
```diff
@@ -297,6 +298,10 @@ func ParseSetting(secrets, volCtx map[string]string, options []string, usePod bo
 		jfsSetting.Envs = env
 	}
 
+	if secrets["extraPreStopHookCmd"] != "" {
+		jfsSetting.ExtraPreStopHookCmd = secrets["extraPreStopHookCmd"]
```
timfeirg: can we simply use `prestop`?

the thing is ... we already have 2 types of naming conventions in this json struct (`secret-key|envs_map`), can't help to feel reluctant to add a third `extraPreStopHookCmd`

pkg/juicefs/mount/builder/common.go
```diff
@@ -96,10 +96,14 @@ func (r *BaseBuilder) genCommonJuicePod(cnGen func() corev1.Container) *corev1.P
 		},
 	}}
 	pod.Spec.Containers[0].Resources = r.jfsSetting.Resources
+	preStopHookCmd := fmt.Sprintf(
+		"umount %s -l; rmdir %s; exit 0", r.jfsSetting.MountPath, r.jfsSetting.MountPath)
+	if r.jfsSetting.ExtraPreStopHookCmd != "" {
+		preStopHookCmd = fmt.Sprintf("%s; %s", r.jfsSetting.ExtraPreStopHookCmd, preStopHookCmd)
```
timfeirg, line 102: strip the trailing `;` from `r.jfsSetting.ExtraPreStopHookCmd`, if user happens to write something like `echo 1;`, the resulting double `;;` will cause error

pkg/config/setting.go
zxh326: what about `extra-prestop-cmd`

user needs to be know that this is an **extra** cmd and not an overwrite?

pkg/juicefs/mount/builder/pod_test.go
```diff
@@ -356,6 +361,16 @@ func TestNewMountPod(t *testing.T) {
 			},
 			want: podMetricTest,
 		},
+		{
+			name: "test-extre-pre-stop-hook-cmd",
+			args: args{
+				name:                "test",
+				cmd:                 defaultCmd,
+				mountPath:           defaultMountPath,
+				extraPreStopHookCmd: "sleep 30",
```
timfeirg: consider using somewhat complex example like `sh -c "cd /tmp && ls; echo 1"`

pkg/config/setting.go
timfeirg: > user needs to be know that this is an extra cmd and not an overwrite?

this is a bit like helm values, where all fields are in fact "extra", but they do not have the "extra" word as prefix, like `annotations`, `labels`

we'll introduce the minor detail in the docs, but even if users misunderstood and treat it like an overwrite, it won't hurt anything, correct?

pkg/config/setting.go
```diff
@@ -56,14 +56,15 @@ type JfsSetting struct {
...
-	Configs       map[string]string `json:"configs_map,omitempty"`
+	SecretKey           string            `json:"secret-key,omitempty"`
+	SecretKey2          string            `json:"secret-key2,omitempty"`
+	Token               string            `json:"token,omitempty"`
+	Passphrase          string            `json:"passphrase,omitempty"`
+	Envs                map[string]string `json:"envs_map,omitempty"`
+	EncryptRsaKey       string            `json:"encrypt_rsa_key,omitempty"`
+	InitConfig          string            `json:"initconfig,omitempty"`
+	Configs             map[string]string `json:"configs_map,omitempty"`
+	ExtraPreStopHookCmd string            `json:"-,omitempty"`
```
timfeirg, line 67: field name `prestop-cmd`?

timfeirg (approved):

PR #913, opened 2024-03-27
Fix panic issue in `IsJobShouldBeRecycled` function

zxh326:
When job does not succeed, completionTime is nil

/cc @timfeirg

### DIFF
--- a/pkg/util/job.go
+++ b/pkg/util/job.go
@@ -42,7 +42,7 @@ func IsJobFailed(job *batchv1.Job) bool {
 }
 
 func IsJobShouldBeRecycled(job *batchv1.Job) bool {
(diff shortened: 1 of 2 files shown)


pkg/util/job.go
```diff
@@ -42,18 +42,21 @@ func IsJobFailed(job *batchv1.Job) bool {
 }
 
 func IsJobShouldBeRecycled(job *batchv1.Job) bool {
-	// job not completed or failed, should not be recycled
+	// job not completed or not failed, should not be recycled
```
timfeirg, line 45: maybe remove the so very obvious comments?

timfeirg (approved):

PR #924, opened 2024-04-08
fix: faster auto recover mountpoint

### LINKED ISSUE #917: [FEATURE] faster automatic recovery

current case:
1. create pod with mountpod
2. delete mountpod
3. node pod
    a. found deletion
    b. recreate mountpod
    c. check the mountpoint. if it is not ready (mountpod have not finished mount), node pod will not try to do recovery, until next checkpoint (10min later)

so mountpoint is not available during `mountpoint start but not ready` to `next check (10min)`

expected: faster recovery

zxh326:
when mountpod recreated/restart, it is possible that the mountpoint may not be ready yet. At this time, the recovery will be skipped and will wait for the next reconciler (now is 10m)

needed to continuously monitor the mount point when rebuilding the mountpod, and as soon as it is ready, immediately recover

fix https://github.com/juicedata/juicefs-csi-driver/issues/917

### DIFF
--- a/pkg/util/pod.go
+++ b/pkg/util/pod.go
@@ -20,8 +20,11 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"os"
 	"strings"
 	"sync"
+	"syscall"
+	"time"
 
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/types"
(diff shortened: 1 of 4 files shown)


pkg/util/pod.go
```diff
@@ -224,3 +227,35 @@ func GetAllRefKeys(pod corev1.Pod) map[string]string {
...
+			time.Sleep(time.Millisecond * 500)
+			continue
+		}
+		if st, ok := finfo.Sys().(*syscall.Stat_t); ok {
+			if st.Ino == 1 {
+				MountPointDevMinorTable.Store(mntPath, DevMinor(st.Dev))
+				klog.V(5).Infof("Mount point %v is ready, mountpod: %s", mntPath, podName)
+				return nil
+			}
+			klog.V(6).Infof("Mount point %v is not ready, mountpod: %s", mntPath, podName)
```
timfeirg, line 255: use the same expression as L245

timfeirg:
I seem to remember that this issue is to be fixed by introducing mount pod readiness probe? I see no related code to add the probe into mount pod?

zxh326:
> I seem to remember that this issue is to be fixed by introducing mount pod readiness probe? I see no related code to add the probe into mount pod?

https://github.com/juicedata/voc/issues/206#issuecomment-2042230496

timfeirg (approved):

PR #926, opened 2024-04-12
fix: delete mountpod when it enter the completed state

zxh326:
when a mountpod enter completed state, we need to consider this an abnormal event that needs to be deleted and enter the recovery process


zxh326:
TBD

PR #937, opened 2024-04-25
fix delete volume issues

zxh326:
- add locks for delete volume

- remove nodeName & priorityClass for job

### DIFF
--- a/pkg/juicefs/mount/builder/job.go
+++ b/pkg/juicefs/mount/builder/job.go
@@ -91,6 +91,10 @@ func (r *JobBuilder) newJob(jobName string) *batchv1.Job {
 			Exec: &corev1.ExecAction{Command: []string{"sh", "-c", "umount /mnt/jfs -l && rmdir /mnt/jfs"}},
 		},
 	}
+	// set node name to empty to let k8s scheduler to choose a node
+	podTemplate.Spec.NodeName = ""
+	// set priority class name to empty to make job use default priority class
+	podTemplate.Spec.PriorityClassName = ""
 	podTemplate.Spec.RestartPolicy = corev1.RestartPolicyOnFailure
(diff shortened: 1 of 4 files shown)


pkg/util/volume.go
```diff
@@ -102,3 +103,29 @@ func getVol(ctx context.Context, client *k8sclient.K8sClient, pod *corev1.Pod, n
...
+type VolumeLocks struct {
+	locks sync.Map
+	mux   sync.Mutex
+}
+
+func NewVolumeLocks() *VolumeLocks {
+	return &VolumeLocks{}
+}
+
+func (vl *VolumeLocks) TryAcquire(volumeID string) bool {
```
timfeirg, line 116: should we introduce timeout for these type of locks? could they cause deadlock?

pkg/juicefs/mount/builder/job.go
```diff
@@ -91,6 +91,10 @@ func (r *JobBuilder) newJob(jobName string) *batchv1.Job {
 			Exec: &corev1.ExecAction{Command: []string{"sh", "-c", "umount /mnt/jfs -l && rmdir /mnt/jfs"}},
 		},
 	}
+	// set node name to empty to let k8s scheduler to choose a node
+	podTemplate.Spec.NodeName = ""
+	// set priority class name to empty to make job use default priority class
+	podTemplate.Spec.PriorityClassName = ""
```
zxh326, line 97: large number of cleanup/delvol jobs may preempt normal/application pods, which is not as expected

pkg/util/volume.go
zxh326: this lock is not waiting, so it's unnecessary

timfeirg (approved):

PR #970, opened 2024-06-07
feat: apply config when recreated mountpod

zxh326:


pkg/controller/pod_driver.go
```diff
@@ -699,29 +699,26 @@ func (p *PodDriver) CleanUpCache(ctx context.Context, pod *corev1.Pod) {
...
-	cpuRequest := pvc.Annotations[config.MountPodCpuRequestKey]
-	memoryRequest := pvc.Annotations[config.MountPodMemRequestKey]
-
-	resources, err := config.ParsePodResources(cpuLimit, memoryLimit, cpuRequest, memoryRequest)
-	if err != nil {
-		return fmt.Errorf("parse pvc resources error: %v", err)
-	}
-	pod.Spec.Containers[0].Resources = resources
+	// update pod spec
+	pod.Labels = attr.Labels
```
timfeirg, line 708: does this mean, all below fields can be customized in configmap, and updated dynamically? (only need to re-create mount pod)

zxh326: yep

timfeirg (approved):

zwwhdls (approved):

Issue #996, opened 2024-06-21
[FEATURE] Support to set nodeSelector for volume job


kyungwan-nam:
**What feature you'd like to add:**

I'd like to set the nodeSelector for volume job.
It looks good if set on the global config introduced https://github.com/juicedata/juicefs-csi-driver/pull/938 

**Why is this feature needed:**

Creating/deleting volume job will be launched on a node where the scheduler chooses by https://github.com/juicedata/juicefs-csi-driver/pull/937

in my cluster, the csi-node is deployed ``juicefs.com/csi-node: "true"`` labeled node only. therefore, the volume job can fail.

zxh326:
may be better to make job NodeSelector/Affinity/Tolerations follow the csi-node configuration

sorry i missed in https://github.com/juicedata/juicefs-csi-driver/pull/937

kyungwan-nam:
@zxh326 
I fixed it as your comment https://github.com/juicedata/juicefs-csi-driver/pull/1003
It looks better. Thanks.

zxh326:
fixed in https://github.com/juicedata/juicefs-csi-driver/pull/1003

PR #997, opened 2024-06-21
Support to set nodeSelector for volume job

kyungwan-nam:
PR for https://github.com/juicedata/juicefs-csi-driver/issues/996


zxh326:
no needed for now

PR #1013, opened 2024-07-03
docs: add delete delay & clean cache example config

zxh326:


timfeirg:
I will take over this pr, this section needs further improvement: https://juicefs.com/docs/csi/guide/resource-optimization/#delayed-mount-pod-deletion

zwwhdls (approved):

Issue #1015, opened 2024-07-03
[FEATURE] Support to keep the volume if Delete reclaim policy


kyungwan-nam:
**What feature you'd like to add:**

Currently, ``Delete`` reclaim policy triggers deleting volume whenever the PVC is deleted.
It looks good if I can decide whether to keep the volume or not. e.g. ``juicefs/keep-volume="true"``

**Why is this feature needed:**

I’d like to keep my data even if the PVC is deleted.
I set the storage class with ``Retain`` reclaim policy. But, when the PVC is deleted, the PV still exists with "released" status.
https://kubernetes.io/docs/concepts/storage/persistent-volumes/#retain
the cluster-admin has to delete the PV with "released" manually. It’s annoying to the cluster-admin.
so, I want to set ``Delete`` reclaim policy with an option to keep the volume.

kyungwan-nam:
Skip deleting volume if ``retainVolume: true`` https://github.com/juicedata/juicefs-csi-driver/pull/1017

zxh326:
https://github.com/juicedata/juicefs-csi-driver/pull/1017#issuecomment-2216897336

PR #1017, opened 2024-07-03
Support to retain the volume if Delete reclaim policy

kyungwan-nam:
PR for https://github.com/juicedata/juicefs-csi-driver/issues/1015

### DIFF
--- a/pkg/driver/provisioner.go
+++ b/pkg/driver/provisioner.go
@@ -187,7 +187,9 @@ func (j *provisionerService) Provision(ctx context.Context, options provisioncon
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


zwwhdls:
It seems strange to retain volume in delete reclaim policy. I don't want to support such a feature, which violates the reclaim semantics of PV. What do you think? cc @zxh326

zxh326:
we don't want to support this feature because it violates the design of CSI

thx..i will close this PR

kyungwan-nam:
Thank you for reviewing it.
I got it. I can handle it in another way.
Thanks.

tsic404:
> Thank you for reviewing it. I got it. I can handle it in another way. Thanks.

How did you deal with this in the end?

PR #1029, opened 2024-07-09
support pass fuse fd to mount pod

zwwhdls:
1. CSI serve `fuse_fd_csi_comm.sock` after create mount pod:
     1. waiting mount pod for sending fuse fd.
     2. provide fuse fd to the mount pod after it restarts.
2. CSI set `JFS_SUPER_COMM` as env in mount pod.
3. When CSI restarted, parse fuse fds of all mount pods  in `/tmp/<mountPodHashVal>`.
4. Stop `fuse_fd_csi_comm.sock` after deleting mount pod.
5. Only do the above things if image tag > v1.2.1 for ce and > 5.1.0 for ee.
6. Do not set `umount` in preStop if this feature opens.
7. If mount pod restart failed, close fuse fd and restart it again.


pkg/controller/pod_driver.go
```diff
@@ -495,6 +506,10 @@ func (p *PodDriver) podReadyHandler(ctx context.Context, pod *corev1.Pod) error
 		return nil
 	}
 
+	if util.ParseClientVersion(pod.Spec.Containers[0].Image).SupportFusePass() {
+		return nil
+	}
```
zxh326: If return here, cannot be store dev minor

pkg/fuse/passfd.go
```diff
@@ -0,0 +1,329 @@
...
+
+func (fs *Fds) GetFdAddress(podHashVal string) string {
+	if f, ok := fs.fds[podHashVal]; ok {
+		return f.serverAddressInPod
+	}
+
+	address := path.Join("/tmp", podHashVal, "fuse_fd_csi_comm.sock")
+	addressInPod := path.Join("/tmp", "fuse_fd_csi_comm.sock")
+	// mkdir parent
+	_ = os.MkdirAll(path.Join("/tmp", podHashVal), 0777)
```
zxh326: this directory doesn't seem to be remove?

pkg/controller/pod_driver.go
```diff
@@ -195,6 +196,10 @@ func (p *PodDriver) checkAnnotations(ctx context.Context, pod *corev1.Pod) error
 				klog.Errorf("Delete pod %s error: %v", pod.Name, err)
 				return err
 			}
+			// close socket
+			if util.ParseClientVersion(pod.Spec.Containers[0].Image).SupportFusePass() {
```
zxh326: can we extract this as a method?

`util.SupportFusePass(pod)`

zxh326: also, if the csi-driver does not mount the `jfs-fuse-fd` hostPath, can it be considered as not supported?

pkg/controller/pod_driver.go
zwwhdls: fixed

pkg/controller/pod_driver.go
zwwhdls: The `jfs-fuse-fd` hostPath will be added in installation yaml and chart of new version by default. It can be ignored.

pkg/fuse/passfd.go
zwwhdls: fixed

pkg/controller/pod_driver.go
```diff
@@ -400,14 +400,15 @@ func (p *PodDriver) podDeletedHandler(ctx context.Context, pod *corev1.Pod) erro
 					klog.Errorf("[podDeletedHandler] Create pod:%s err:%v", pod.Name, err)
 				}
 
-				if util.ParseClientVersion(pod.Spec.Containers[0].Image).SupportFusePass() {
-					return nil
-				}
-
```
zxh326: need fix `podReadyHandler`

pkg/controller/pod_driver.go
zxh326: worried about user only updated the image version

zxh326 (approved):

Issue #1022, opened 2024-07-05
performance optimization


zxh326:
This issue is used to track some performance issues

## reduce k8s api call

- list all apis
	- [x] 	[pod_controller.go#77](https://github.com/juicedata/juicefs-csi-driver/blob/2706a13a96aec04ed432dfd444828d272877fc40/pkg/controller/pod_controller.go#L77)
	- [x] 	~~[pv webhook](https://github.com/juicedata/juicefs-csi-driver/blob/2706a13a96aec04ed432dfd444828d272877fc40/pkg/webhook/handler/handler.go#L179)~~
	   only list in create static PV
	- [x] 	~~[GetNamespace](https://github.com/juicedata/juicefs-csi-driver/blob/2706a13a96aec04ed432dfd444828d272877fc40/pkg/util/resource/namespace.go#L47)~~
	   only list in sidecar mode, when app pod namespace is nil
	- [ ] [check for subpath](https://github.com/juicedata/juicefs-csi-driver/blob/2706a13a96aec04ed432dfd444828d272877fc40/pkg/util/resource/pvc.go#L110)
	- [ ] [check for secret finalizer](https://github.com/juicedata/juicefs-csi-driver/blob/2706a13a96aec04ed432dfd444828d272877fc40/pkg/util/resource/pvc.go#L157)
		need refactor, see https://github.com/juicedata/juicefs-csi-driver/issues/1023
    - [x] unnecessary api request in different node
    	fixed https://github.com/juicedata/juicefs-csi-driver/pull/1047 
	
- [ ] [multi controller runtime manager in controller](https://github.com/juicedata/juicefs-csi-driver/blob/53a87aa57015628decf11f9c9136da135c8023ec/cmd/controller.go#L159-L186)

PR #1031, opened 2024-07-10
chore: add PV subpath label to provisioned PV

zxh326:
- add a label `subPath` to the PV object. this change allows for better tracking of pv with specific subPath, maybe useful in the future, but currently useless

ref: https://github.com/juicedata/juicefs-csi-driver/issues/1022

### DIFF
--- a/pkg/util/resource/pvc.go
+++ b/pkg/util/resource/pvc.go
@@ -27,6 +27,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 
 	v1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/klog"
 
 	"github.com/juicedata/juicefs-csi-driver/pkg/config"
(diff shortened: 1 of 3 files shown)


pkg/util/resource/pvc.go
```diff
@@ -100,19 +100,30 @@ func CheckForSubPath(ctx context.Context, client *k8s.K8sClient, volume *v1.Pers
...
 
-	if sc == "" {
+	if scName == "" {
 		return false, nil
 	}
 
+	sc, err := client.GetStorageClass(ctx, scName)
+	if err == nil {
+		// if the storage class does not set the pathPattern, subpath used unique pv name, should delete
+		if _, ok := sc.Parameters["pathPattern"]; !ok {
```
zwwhdls: Did [`pathPattern == ""`](https://github.com/juicedata/juicefs-csi-driver/blob/b560adbb9becee15c15247eeb887cc71950e24d4/pkg/util/resource/pvc.go#L99) did the same thing? 

If pv.spec.csi.volumeAttributes (copy from sc) has no pathPattern, do not check other pvs for subPath.

zxh326: you are right, I missed it

zwwhdls (approved):

PR #1047, opened 2024-07-22
Add mountpod filter in mount manager predicates 

zxh326:
- add mountpod filter in mount manager predicates to avoid unnecessary api request
- improve error logging for mountpod fuse connection

### DIFF
--- a/pkg/controller/pod_controller.go
+++ b/pkg/controller/pod_controller.go
@@ -130,11 +130,17 @@ func (m *PodController) SetupWithManager(mgr ctrl.Manager) error {
 	return c.Watch(&source.Kind{Type: &corev1.Pod{}}, &handler.EnqueueRequestForObject{}, predicate.Funcs{
 		CreateFunc: func(event event.CreateEvent) bool {
 			pod := event.Object.(*corev1.Pod)
+			if pod.Spec.NodeName != config.NodeName && pod.Spec.NodeSelector["kubernetes.io/hostname"] != config.NodeName {
+				return false
+			}
 			klog.V(6).Infof("watch pod %s created", pod.GetName())
 			return true
 		},
 		UpdateFunc: func(updateEvent event.UpdateEvent) bool {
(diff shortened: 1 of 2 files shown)


pkg/controller/pod_controller.go
```diff
@@ -130,11 +130,17 @@ func (m *PodController) SetupWithManager(mgr ctrl.Manager) error {
 	return c.Watch(&source.Kind{Type: &corev1.Pod{}}, &handler.EnqueueRequestForObject{}, predicate.Funcs{
 		CreateFunc: func(event event.CreateEvent) bool {
 			pod := event.Object.(*corev1.Pod)
+			if pod.Spec.NodeName != config.NodeName && pod.Spec.NodeSelector["kubernetes.io/hostname"] != config.NodeName {
```
zwwhdls, line 133: && or || ?

zxh326: &&

zxh326: copy from https://github.com/juicedata/juicefs-csi-driver/blob/3e81a8d505dab116bee9504d2fde1cba6eb73e6c/pkg/controller/pod_controller.go#L62-L65

zwwhdls: I think it should be "||", because mount pod may not have nodeSelector. The line in reconcile is wrong too, please correct it as well, thx.

zxh326: `||` is wrong

example:


```

config.NodeName = k8s-01

# mountpod
nodeName: node-01
nodeSelector: nil
```

`mountPod.Spec.NodeName != config.NodeName` => `false`

`mountPod.Spec.NodeSelector["kubernetes.io/hostname"] != config.NodeName` => `true`

then mount pod will be skipped reconciler

zwwhdls: ok, u r right.

zwwhdls (approved):

PR #1073, opened 2024-08-05
feat: support custom mountpod volumes

zxh326:
user can add some volumes to mountpod

like this
```yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: juicefs-csi-driver-config
  namespace: kube-system
data:
  config.yaml: |-
    mountPodPatch:    
    - pvcSelector:
        matchLabels:
          need-block-device: "true"
      volumeDevices:
        - name: block-devices
          devicePath: /dev/sda1
      volumes:
        - name: block-devices
          persistentVolumeClaim:
            claimName: block-pvc
```

ref: https://github.com/juicedata/voc/issues/270#issuecomment-2268088105

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -138,6 +138,24 @@ const (
 	DefaultMountPodMemRequest = "1Gi"
 )
 
+var interVolumesPrefix = []string{
+	"rsa-key",
+	"init-config",
+	"config-",
+	"jfs-dir",
+	"update-db",
+	"cachedir-",
+}
+
+func IsInterVolume(name string) bool {
+	for _, prefix := range interVolumesPrefix {
+		if strings.HasPrefix(name, prefix) {
+			return true
+		}
+	}
+	return false
+}
+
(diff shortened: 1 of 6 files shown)


pkg/config/config.go
```diff
@@ -250,6 +272,58 @@ func (mpp *MountPodPatch) merge(mp MountPodPatch) {
...
+			}
+			vok[v.Name] = true
+			mpp.Volumes = append(mpp.Volumes, v)
+		}
+	}
+	if mp.VolumeMounts != nil {
+		if mpp.VolumeMounts == nil {
+			mpp.VolumeMounts = []corev1.VolumeMount{}
+		}
+		for _, vm := range mp.VolumeMounts {
```
zwwhdls, line 304: Does `volumeMounts` also need to check interVolume?

zxh326: no need, it will check the exist of the used volume, which has already been check interVolume

zwwhdls (approved):

PR #1092, opened 2024-08-21
fix: clear old secret data when update secret

zxh326:


pkg/juicefs/juicefs.go
```diff
@@ -413,6 +413,7 @@ func (j *juicefs) genJfsSettings(ctx context.Context, volumeID string, target st
 	}
 	klog.V(6).Infof("Get uniqueId of volume [%s]: %s", volumeID, uniqueId)
 	jfsSetting.UniqueId = uniqueId
+	jfsSetting.SecretName = fmt.Sprintf("juicefs-%s-secret", jfsSetting.UniqueId)
```
zxh326, line 416: make the secret name calculate hash

zxh326:
先合

PR #1097, opened 2024-08-27
refact log

zwwhdls:


zxh326 (approved):
now `--v=1` mean debug log?

also need update helm chart

https://github.com/juicedata/charts/blob/6110e266310f4180f282c661474be46a18a282a9/charts/juicefs-csi-driver/templates/daemonset.yaml#L57-L61

https://github.com/juicedata/charts/blob/6110e266310f4180f282c661474be46a18a282a9/charts/juicefs-csi-driver/templates/controller.yaml#L55-L59

zwwhdls:
will update chart after release it.

PR #1103, opened 2024-09-03
support mountOptions in configmap

zxh326:


pkg/config/setting.go
```diff
@@ -728,6 +732,19 @@ func applyAttrPatch(attr *PodAttr, setting *JfsSetting) {
 	attr.VolumeMounts = patch.VolumeMounts
 	attr.Volumes = patch.Volumes
 	attr.Env = patch.Env
+
+	// merge options
+	if len(setting.Options) > 0 && len(patch.MountOptions) > 0 {
```
zwwhdls: what if len(setting.Options) == 0 and len(patch.MountOpions) > 0 ?

zxh326: fixed

pkg/config/setting.go
```diff
@@ -728,6 +732,19 @@ func applyAttrPatch(attr *PodAttr, setting *JfsSetting) {
 	attr.VolumeMounts = patch.VolumeMounts
 	attr.Volumes = patch.Volumes
 	attr.Env = patch.Env
+
+	// merge or overwrite setting options
+	if setting.Options == nil {
+		setting.Options = make([]string, 0)
+	}
+	for _, option := range patch.MountOptions {
```
zwwhdls, line 740: if patch.MountOptions is nil?

zxh326: for loop will not execute

zwwhdls (approved):

PR #1122, opened 2024-09-23
fix apply config in `STORAGE_CLASS_SHARE_MOUNT` mode

### LINKED ISSUE #1065: [BUG] restart mountpod cannot apply config in `STORAGE_CLASS_SHARE_MOUNT` mode
because it is impossible to obtain the mountpod pv

zxh326:
fix: https://github.com/juicedata/juicefs-csi-driver/issues/1065


pkg/config/setting.go
```diff
@@ -757,3 +769,17 @@ func IsCEMountPod(pod *corev1.Pod) bool {
 	}
 	return false
 }
+
+func getPVNameFromTarget(target string) string {
```
zwwhdls, line 774: Add unit test for this function

zwwhdls (approved):

PR #1143, opened 2024-10-12
fix concurrency issue when upgrade

### LINKED ISSUE #1145: [BUG] 延迟删除的mount pod平滑升级后立刻被删除
**What happened**:
1、配置mount pod   juicefs/mount-delete-delay: 3h ,创建应用pod
2、等应用pod  completed后进行平滑升级
3、mount pod 被删除，预期不应该被删除

**Environment**:
- Others:

zwwhdls:
1. select no complete mount pod when mount/umount
2. remove finalizer until new mount pod created successfully when old is deleted
3. send SIGHUP only once when upgrade
4. check old mount pod support smooth upgrade 
5. fix https://github.com/juicedata/juicefs-csi-driver/issues/1145

### DIFF
--- a/pkg/controller/pod_driver.go
+++ b/pkg/controller/pod_driver.go
@@ -246,11 +246,11 @@ func (p *PodDriver) podCompleteHandler(ctx context.Context, pod *corev1.Pod) (Re
 	lock.Lock()
 	defer lock.Unlock()
 
-	needCreate, err := p.needCreateMountPod(ctx, pod.Labels[common.PodUniqueIdLabelKey], hashVal)
+	notAvailable, err := p.hasAvailableMountPod(ctx, pod.Labels[common.PodUniqueIdLabelKey], hashVal)
 	if err != nil {
 		return Result{}, err
 	}
-	if needCreate {
+	if notAvailable {
 		newPodName := podmount.GenPodNameByUniqueId(pod.Labels[common.PodUniqueIdLabelKey], true)
 		log.Info("need to create a new one", "newPodName", newPodName)
 		newPod, err := p.newMountPod(ctx, pod, newPodName)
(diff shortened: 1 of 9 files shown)


pkg/controller/pod_driver.go
```diff
@@ -438,49 +450,42 @@ func (p *PodDriver) podDeletedHandler(ctx context.Context, pod *corev1.Pod) (Res
...
-			newPod, err := p.newMountPod(ctx, pod, newPodName)
-			if err == nil {
-				_, err = p.Client.CreatePod(ctx, newPod)
-				if err != nil {
-					log.Error(err, "Create pod")
-				}
+	if notAvailable {
+		// create pod
+		newPodName := podmount.GenPodNameByUniqueId(pod.Labels[common.PodUniqueIdLabelKey], true)
+		log.Info("pod targetPath not empty, need to create a new one", "newPodName", newPodName)
```
zxh326: why not available mount pods mean **pod targetPath not empty**?

pkg/controller/pod_driver.go
```diff
@@ -968,13 +973,13 @@ func (p *PodDriver) needCreateMountPod(ctx context.Context, uniqueId, hashVal st
...
-	needCreate := true
+	notAvailable := true
 	for _, po := range pods {
 		if po.DeletionTimestamp == nil && !resource.IsPodComplete(&po) {
-			needCreate = false
+			notAvailable = false
 		}
 	}
-	return needCreate, nil
+	return notAvailable, nil
```
zxh326: func name is has available mount, but return not available pod, It's very confusing.

pkg/controller/pod_driver.go
zxh326: How to manually delete a pod in a delayed deletion state(with no refs)

zwwhdls: not available means there are no available mount pod in cluster. If pod targetPath not empty and no available , need to create a new one.

zwwhdls: If manually delete the delay deleted pod, it will not be handled as we discussed before.

zxh326: If I manually delete a delay pod, will a new one be created? even if there's no reference?

zxh326: > not available means there are no available mount pod in cluster. If pod targetPath not empty and no available , need to create a new one.

should we check `len(existTargets) > 0` here?

check `len(existTargets) == 0` before but with other condition..

https://github.com/juicedata/juicefs-csi-driver/blob/895ddad32d50e0325fb469aec1a6f97a12ed17e8/pkg/controller/pod_driver.go#L435

zxh326 (approved):

PR #1148, opened 2024-10-18
fix: use pod name to get lock if hashVal is nil

### LINKED ISSUE #1147: [BUG] csi v0.25.0 can not manage exist mount pod which has no hashval annotation
**What happened**:
In version v0.25.0, a new mount pod annotation `hashval` was added for passing fusefd to mount pod. I check the pod driver code, found it will throw error, if mount pod has no `hashval` annotation(create by old version csi).
https://github.com/juicedata/juicefs-csi-driver/blob/c545bcd8d343d10eca9cd49959edb1dac52b596a/pkg/controller/pod_driver.go#L152-L158

And worse, csi uses err group to handle mount pod. since exist mount pod `checkAnnotations` throw a error, this driver will cancel other mount pods' reconcile.

https://github.com/juicedata/juicefs-csi-driver/blob/c545bcd8d343d10eca9cd49959edb1dac52b596a/pkg/controller/reconciler.go#L147-L153

remove `hashval` annotaintion check, not throw an error for exist mount pod.

* have mount pod create by 0.24.x
* upgrade csi to v0.25.0

zwwhdls:
fix https://github.com/juicedata/juicefs-csi-driver/issues/1147

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -35,6 +35,7 @@ import (
 
 	corev1 "k8s.io/api/core/v1"
 
+	"github.com/juicedata/juicefs-csi-driver/pkg/common"
 	k8s "github.com/juicedata/juicefs-csi-driver/pkg/k8sclient"
 )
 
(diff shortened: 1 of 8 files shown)


SonglinLife:
Perhaps the `StopFd` should include additional code to handle an empty hashVal. If hashVal is an empty string, StopFd will remove the fs.basePath.
https://github.com/juicedata/juicefs-csi-driver/blob/c545bcd8d343d10eca9cd49959edb1dac52b596a/pkg/fuse/passfd/passfd.go#L139-L147

zxh326 (approved):

PR #1160, opened 2024-10-29
fix apply mountOptions

zxh326:
- fix overwrite exist mountOptions
- if `buffer-size` is greater than pod memory limit, fallback to memory limit

### DIFF
--- a/pkg/config/setting_test.go
+++ b/pkg/config/setting_test.go
@@ -26,6 +26,7 @@ import (
 	"k8s.io/apimachinery/pkg/api/resource"
 
 	"github.com/juicedata/juicefs-csi-driver/pkg/common"
+	"github.com/stretchr/testify/assert"
 )
 
 var (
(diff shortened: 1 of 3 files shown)


pkg/config/setting_test.go
```diff
@@ -1123,3 +1124,78 @@ func Test_getPVNameFromTarget(t *testing.T) {
...
+			want: &JfsSetting{
+				Attr: &PodAttr{
+					Resources: corev1.ResourceRequirements{
+						Limits: map[corev1.ResourceName]resource.Quantity{
+							corev1.ResourceMemory: resource.MustParse("10Mi"),
+						},
+					},
+				},
+				Options: []string{
+					"buffer-size=10Mi",
```
zxh326:
```suggestion
					"buffer-size=10485760",
```

zwwhdls (approved):

PR #1173, opened 2024-11-11
fix: delete reference in old mount pod when republish, fix #1172

### LINKED ISSUE #1172: [BUG] Old Mount Pod is leak when config is wrong
**What happened**:
1. Config is wrong and old mount pod won't be ready. Correct the config and CSI Node will recreate a new pod util it is ready
2. But kubelet will retry while Publish call receives error. And a new Mount Pod will be create it with new config.
3. Both two mount pods are ready in the cluster, but only the new Mount Pod is in serivce.

**Environment**:
- Others:

zwwhdls:
fix #1172

### DIFF
--- a/pkg/juicefs/mount/pod_mount.go
+++ b/pkg/juicefs/mount/pod_mount.go
@@ -324,6 +324,9 @@ func (p *PodMount) genMountPodName(ctx context.Context, jfsSetting *jfsConfig.Jf
 	var podName string
 	for _, pod := range pods {
 		po := pod
+		if pod.Spec.NodeName != jfsConfig.NodeName && pod.Spec.NodeSelector["kubernetes.io/hostname"] != jfsConfig.NodeName {
+			continue
+		}
 		if po.Labels[common.PodJuiceHashLabelKey] != jfsSetting.HashVal {
 			for k, v := range pod.Annotations {
 				if v == jfsSetting.TargetPath {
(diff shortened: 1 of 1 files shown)


pkg/juicefs/mount/pod_mount.go
```diff
@@ -314,23 +313,38 @@ func (p *PodMount) JDeleteVolume(ctx context.Context, jfsSetting *jfsConfig.JfsS
...
 	}}
 	pods, err := p.K8sClient.ListPod(ctx, jfsConfig.Namespace, labelSelector, nil)
 	if err != nil {
 		log.Error(err, "List pods of uniqueId", "uniqueId", jfsSetting.UniqueId, "hashVal", jfsSetting.HashVal)
 		return "", err
 	}
+	var podName string
 	for _, pod := range pods {
+		po := pod
+		if po.Labels[common.PodJuiceHashLabelKey] != jfsSetting.HashVal {
```
zxh326, line 330: check node?

pkg/juicefs/mount/pod_mount.go
```diff
@@ -314,23 +313,38 @@ func (p *PodMount) JDeleteVolume(ctx context.Context, jfsSetting *jfsConfig.JfsS
...
+						return "", err
+					}
+				}
+			}
+			continue
+		}
 		if pod.DeletionTimestamp != nil || resource.IsPodComplete(&pod) {
 			continue
 		}
 		if pod.Spec.NodeName == jfsConfig.NodeName || pod.Spec.NodeSelector["kubernetes.io/hostname"] == jfsConfig.NodeName {
```
zxh326: check hash value?

zwwhdls: checked in https://github.com/juicedata/juicefs-csi-driver/blob/7096422f42978608185ea8fb87cdce34c56113e3/pkg/juicefs/mount/pod_mount.go#L327

and continue here if hash value is not equal https://github.com/juicedata/juicefs-csi-driver/blob/7096422f42978608185ea8fb87cdce34c56113e3/pkg/juicefs/mount/pod_mount.go#L336

zxh326 (approved):

PR #1175, opened 2024-11-12
make batch upgrade run parallel and set pvc selector

zwwhdls:
1. batch upgrade run in parallel
2. set parallel num in dashboard
3. set pvc selector in batch upgrade

CleanShot 2024-12-05 at 18 00 00


pkg/dashboard/batch.go
```diff
@@ -329,7 +342,111 @@ func newUpgradeJob(nodeName string, recreate bool) *batchv1.Job {
...
+	s, _ = metav1.LabelSelectorAsSelector(ls)
+	if err = api.cachedReader.List(ctx, &secretList, &client.ListOptions{LabelSelector: s}); err != nil {
+		return nil, nil, err
+	}
+
+	pvMap := make(map[string]*corev1.PersistentVolume)
+	pvcMap := make(map[string]*corev1.PersistentVolumeClaim)
+	secretMap := make(map[string]*corev1.Secret)
+	for _, pv := range pvs {
+		pv2 := pv
```
zxh326: We've upgraded to 1.22, so we don't have to do this.

cmd/dashboard/upgrade.go
```diff
@@ -18,98 +18,174 @@ package main
...
+					resultCh <- u.triggerUpgrade(ctx, &mp2)
+				}()
+			}
+		}()
+
+		for oneErr := range resultCh {
+			if oneErr != nil {
+				if !u.conf.IgnoreError {
+					fmt.Printf("%s BATCH-FAIL some pods upgrade failed\n", time.Now().Format(time.DateTime))
+					return
```
zxh326, line 153: What happens to the other pods that are being upgraded after the return here?

go.mod
```diff
@@ -16,6 +16,8 @@ require (
 	github.com/smartystreets/goconvey v1.6.4
 	github.com/spf13/cobra v1.8.1
 	github.com/stretchr/testify v1.9.0
+	github.com/sykesm/zap-logfmt v0.0.4
+	go.uber.org/zap v1.26.0
```
zxh326: not used

zxh326 (commented):
run `pnpm run format` in the dashboard-ui

cmd/dashboard/upgrade.go
zwwhdls: complete upgrading until fail or success. We can not cancel it.

zxh326 (approved):

Issue #1194, opened 2024-11-27
[FEATURE] use CSI Node pod to get CSI spec instead of DaemonSet


zwwhdls:
**What feature you'd like to add:**
In some special cluster such as alibaba cloud, people may use some CR (such as [Advanced Daemonset](https://openkruise.io/docs/next/user-manuals/advanceddaemonset/)) instead of DaemonSet to achieve more advantages. 

To be more general, we should use list pod by label select instead of get Daemonset in CSI Node.

**Why is this feature needed:**

showjason:
Hi @zwwhdls ,
I'd like to solve this issue, however, I'm curious about the motivation behind this feature request. From my understanding, getting CSI pod specs through a CSI Node pod instead of a DaemonSet seems like it would require an additional step of querying all pods in the target namespace and then filtering based on OwnerReference to find the right CSI Node pods before finally obtaining the pod.Spec.

Could you share the advantages of this approach? Is it related to performance optimization, architectural considerations, or solving particular issues you've encountered with the current DaemonSet implementation?

Or maybe my understanding is incorrect, anyway, please considering my question!
Thanks in advance!

zwwhdls:
Hi @showjason , thanks! I have updated the reason of the issue in description.

Besides, namespace can get from downward api and csi node pods have specific labels. So listing pods by labelSelect can work.

showjason:
Ah, thank you! I hadn't considered using labelSelector to catch the CSI node pods - this is a better approach!
Please assign this feature to me, I will implement it.
/assign

showjason:
Hi @zwwhdls 
Yesterday, I did investigation and test, I found that it should be not feasible to obtain the correct CSI Node pod using labelSelector, because the labels of CSI Node pods are customizable(I think the specific labels you mentioned are not immutable). We cannot predict the labels set by users, so we cannot accurately obtain the CSI Node pod via labels. As of now, there are two approaches I can think of to adapt to CRs like Advanced DaemonSet 
1. Add corresponding methods to obtain the DaemonSet for different CRs, like the existing PR #1197 
2. Retrieve all pods and then find the CSI Node pod vi `ownerReference`. However, if there are a large number of pods in the namespace where the CSI Node pod is located, this is most likely to cause performance issues.

zwwhdls:
[...]

PR #1197, opened 2024-11-30
support juicefs-csi-node run as ads

LiuYangkuan:
#1194 
[FEATURE] use CSI Node pod to get CSI spec instead of DaemonSet


zwwhdls (changes requested):
Please do not use non-native resources and use list CSI Node pods instead, as the issue described.  @LiuYangkuan

zxh326:
closed because it is a duplicate of https://github.com/juicedata/juicefs-csi-driver/pull/1348

PR #1209, opened 2024-12-16
make batch upgrade multi jobs

zwwhdls:
1. support multi jobs of batch upgrade
2. support job pause, resume and stop

CleanShot 2024-12-16 at 15 34 01

CleanShot 2024-12-16 at 15 34 31


cmd/dashboard/upgrade.go
```diff
@@ -60,43 +60,80 @@ var upgradeCmd = &cobra.Command{
...
 
-func init() {
-	upgradeCmd.Flags().StringVar(&batchConfigName, "batchConfigName", "juicefs-upgrade-batch", "configmap name for batch upgrade")
+func (u *BatchUpgrade) handleSignal() {
+	sigChan := make(chan os.Signal, 10)
+	signal.Notify(sigChan, syscall.SIGUSR1, syscall.SIGTERM)
+
+	paused := false
+	for {
+		sig := <-sigChan
```
zxh326:
```suggestion
	for sig := range sigChan {
```

cmd/dashboard/upgrade.go
```diff
@@ -106,75 +143,148 @@ type BatchUpgrade struct {
...
+			wg.Wait()
+			close(resultCh)
+		}()
+		for _, mp := range batch {
+			wg.Add(1)
+			go func(mp config.MountPodUpgrade) {
+				resultCh <- u.triggerUpgrade(ctx, &mp)
+				wg.Done()
+			}(mp)
+		}
```
zxh326, line 233: How does this `Parallel` work?

cmd/dashboard/upgrade.go
```diff
@@ -106,75 +143,148 @@ type BatchUpgrade struct {
...
-			fmt.Printf("%s BATCH-FAIL upgrade timeout\n", time.Now().Format(time.DateTime))
+func (u *BatchUpgrade) processBatch(ctx context.Context) {
+	if u.crtBatch > len(u.conf.Batches) {
+		return
+	}
+	u.setCrtBatchStatus(config.Running)
+	var (
+		wg       sync.WaitGroup
+		resultCh = make(chan error, u.conf.Parallel)
+		batch    = u.conf.Batches[u.crtBatch-1]
```
zxh326:
```suggestion
		batch    = u.conf.Batches[u.crtBatch-1]
		resultCh = make(chan error, len(batch))
```

Is this understanding correct?

cmd/dashboard/upgrade.go
zwwhdls: dashboard will set pods in batch with parallel, which batch has parallel numbers of pods.

cmd/dashboard/upgrade.go
zwwhdls: parallel always equals to length of batch, except the last batch which might be less than parallel

zxh326 (approved):

PR #1213, opened 2024-12-18
fix: only handle used secret in secret-controller

zxh326:
Only handle the currently used secret to perform auth and add initconfig

- add a pv controller to save used secret

### DIFF
--- a/pkg/controller/pv_controller.go
+++ b/pkg/controller/pv_controller.go
@@ -61,7 +61,7 @@ func (m *PVController) Reconcile(ctx context.Context, request reconcile.Request)
 		secretNamespace := pv.Spec.CSI.NodePublishSecretRef.Namespace
 		watchedSecrets[fmt.Sprintf("%s/%s", secretNamespace, secretName)] = struct{}{}
 		// for first time, we need to refresh the secret init config in pv controller
-		if err := refreshSecretInitConfig(ctx, m.K8sClient, secretNamespace, secretName); err != nil {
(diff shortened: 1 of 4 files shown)


pkg/controller/pv_controller.go
```diff
@@ -0,0 +1,125 @@
...
+		secretNamespace := pv.Spec.CSI.NodePublishSecretRef.Namespace
+		if _, ok := watchedSecrets[secretName]; !ok {
+			watchedSecrets[fmt.Sprintf("%s/%s", secretNamespace, secretName)] = struct{}{}
+			sc := NewSecretController(m.K8sClient)
+			_, err := sc.Reconcile(ctx, reconcile.Request{
+				NamespacedName: types.NamespacedName{
+					Namespace: secretNamespace,
+					Name:      secretName,
+				},
+			})
```
zwwhdls: make a function called `RefreshSecret`, and called by both here and secretController.

pkg/controller/pv_controller.go
```diff
@@ -0,0 +1,125 @@
...
+			}
+		}
+	}
+	return reconcile.Result{}, nil
+}
+
+func shouldPVInQueue(pv *corev1.PersistentVolume) bool {
+	if pv.Spec.CSI != nil {
+		return pv.Spec.CSI.Driver == config.DriverName
+	}
```
zwwhdls, line 83: if secret of pv already in `watchedSecret`, do not requeue

zwwhdls (approved):
Other LGTM

PR #1261, opened 2025-02-07
fix set quota

### LINKED ISSUE #1260: [BUG] not return quota timeout error in ControllerExpand api
**What happened**:

In the `ControllerExpand` API, if a quota setting timeout error, we return success and running in background

https://github.com/juicedata/juicefs-csi-driver/blob/2814fcaee5c7ebfe1f418353e566625d991b576e/pkg/juicefs/juicefs.go#L659-L662

If the quota continues to fail in the background, there will be no opportunity to retry, and we need to fix this behavior.

**Environment**:
- JuiceFS CSI Driver version (which image tag did your CSI Driver use): v0.23.1~
- Others:

zxh326:
fix #1260 

- Remove the logic that runs in the background upon `setQuota` timeout in the community version.

  after https://github.com/juicedata/juicefs-csi-driver/pull/1223, `NodePublish api` always running setQuota in background
  
  when expanding PVCs, timeouts must return errors to permit retries.

- Adjust csi-resizer timeout to 20s.

### DIFF
--- a/pkg/juicefs/juicefs.go
+++ b/pkg/juicefs/juicefs.go
@@ -621,7 +621,7 @@ func (j *juicefs) SetQuota(ctx context.Context, secrets map[string]string, jfsSe
 		cmdArgs = []string{config.CliPath, "quota", "set", secrets["name"], "--path", quotaPath, "--capacity", strconv.FormatInt(cap, 10)}
 	}
 	log.Info("quota cmd", "command", strings.Join(cmdArgs, " "))
-	cmdCtx, cmdCancel := context.WithTimeout(ctx, 5*defaultCheckTimeout)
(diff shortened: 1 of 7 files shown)


pkg/juicefs/juicefs.go
```diff
@@ -644,25 +644,14 @@ func (j *juicefs) SetQuota(ctx context.Context, secrets map[string]string, jfsSe
 		return wrapSetQuotaErr(string(res), err)
 	}
 
-	done := make(chan error, 1)
-	go func() {
-		// ce cli will block until quota is set
```
zxh326, line 649: If the meta is abnormal, it won't block here

zwwhdls (approved):
Make sure `timeout` in resizer is also added in chart

PR #1299, opened 2025-03-18
feat: add validation webhook for pod eviction

### LINKED ISSUE #856: [BUG] mount pod terminates before my application pod when draining a node
### **What you expected to happen**:
When either manually draining a node or using node termination handler to automatically drain spot nodes 2 minutes before AWS spot node is taken away, all pods on said node receive SIGTERM.

This includes my application pod(s) and the mount pod.

**A gracefully designed mount pod should delay its termination until my application pod terminates completely,**

since the application pod may handle SIGTERM to perform clean up logic (for its own graceful termination) and might need to use the mount volume during so.

### **What happened**:
Instead, the current outcome of draining a node is that the mount pod does **not** wait for my application pod to terminate first.

The mount pod terminates before my application pod, causing my application to meet `Transport endpoint is not connected` error when performing clean up logic.

### **How to reproduce it (as minimally and precisely as possible)**:
1. Prepare a juicefs PV (either via static provisioning or dynamic provisioning) in a test namespace
2. Deploy a workload like the following:
```yaml
apiVersion: apps/v1
kind: Deployment
metadata:
  name: my-juicefs-test
  namespace: test-namespace
spec:
  selector:
    matchLabels:
      app: my-juicefs-test
  template:
    metadata:
      labels:
        app: my-juicefs-test
    spec:
      containers:
      - image: ubuntu
        args:
        - bash
        - -c
        - 'while true; do ls /mnt/my-mount; sleep 1; done'
        imagePullPolicy: IfNotPresent
        name: my-juicefs-test
        resources:
          limits:
            cpu: 200m
            memory: 128Mi
        volumeMounts:
        - mountPath: /mnt/my-mount
          mountPropagation: HostToContainer
          name: my-mount
          readOnly: true
      terminationGracePeriodSeconds: 300
      volumes:
      - name: my-mount
        persistentVolumeClaim:
          claimName: my-pvc
```
3. Observe that an application pod and a mount pod are created on a node
   - Tip: isolate the node using taints so that no other workloads are deployed to this node that will be used for test purposes.
4. Drain the node
   - Note that the `bash -c` entrypoint command does not propagate SIGTERM. Therefore the application pod container does not terminate right away; instead the while loop keeps going. In effect we are able to simulate a clean up logic that makes use of the juicefs volume mount.
5. Observe that the mount pod terminates disregarding the application pod status. Also observe that the application pod starts logging `Transport endpoint is not connected` errors.

### **Environment**:
- JuiceFS CSI Driver version (which image tag did your CSI Driver use): `helm chart version: 0.19.5`, `image tag version: v0.23.2`
- Kubernetes version (e.g. `kubectl version`): `Server Version: version.Info{Major:"1", Minor:"28+", GitVersion:"v1.28.4-eks-8cb36c9", GitCommit:"b4ddd686b005f445825bd0afd5dba24d4414d4cf", GitTreeState:"clean", BuildDate:"2023-11-22T21:52:14Z", GoVersion:"go1.20.11", Compiler:"gc", Platform:"linux/amd64"}`
- Object storage (cloud provider and region): AWS S3
- Metadata engine info (version, cloud provider managed or self maintained): AWS Elasticache Redis
- Others:

zxh326:
fix: https://github.com/juicedata/juicefs-csi-driver/issues/856

### DIFF
--- a/pkg/webhook/handler/handler.go
+++ b/pkg/webhook/handler/handler.go
@@ -183,3 +183,49 @@ func (s *PVHandler) Handle(ctx context.Context, request admission.Request) admis
 	}
 	return admission.Allowed("")
 }
+
+var (
+	evictLog = klog.NewKlogr().WithName("evict-pod-handler")
+)
+
+type EvictPodHandler struct {
+	client *k8sclient.K8sClient
+	// A decoder will be automatically injected
+	decoder admission.Decoder
+}
+
+func NewEvictPodHandler(client *k8sclient.K8sClient, scheme *runtime.Scheme) *EvictPodHandler {
+	return &EvictPodHandler{
+		client:  client,
+		decoder: admission.NewDecoder(scheme),
+	}
+}
+
+func (s *EvictPodHandler) Handle(ctx context.Context, request admission.Request) admission.Response {
+	if request.Namespace != config.Namespace {
+		return admission.Allowed("")
+	}
+	if request.SubResource != "eviction" || request.Operation != "CREATE" {
+		evictLog.Info("skip evict pod", "request", request)
+		return admission.Allowed("")
+	}
+	evictLog.Info("receive evict pod request", "pod", request.Name, "namespace", request.Namespace)
+	pod, err := s.client.GetPod(ctx, request.Name, request.Namespace)
+	if err != nil {
+		evictLog.Error(err, "get pod failed", "name", request.Name, "namespace", request.Namespace)
+		return admission.Errored(http.StatusBadRequest, err)
+	}
+	if value, ok := pod.Labels[common.PodTypeKey]; !ok || value != common.PodTypeValue {
+		evictLog.Info("skip evict pod because it's not a mount pod", "name", request.Name, "namespace", request.Namespace)
+		return admission.Allowed("")
+	}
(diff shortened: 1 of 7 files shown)


pkg/webhook/handler/handler.go
```diff
@@ -183,3 +183,49 @@ func (s *PVHandler) Handle(ctx context.Context, request admission.Request) admis
...
+	}
+	if value, ok := pod.Labels[common.PodTypeKey]; !ok || value != common.PodTypeValue {
+		evictLog.Info("skip evict pod because it's not a mount pod", "name", request.Name, "namespace", request.Namespace)
+		return admission.Allowed("")
+	}
+
+	for k, target := range pod.Annotations {
+		if k == util.GetReferenceKey(target) {
+			evictLog.Info("deny evict mount pod because it's has juicefs reference", "name", request.Name, "namespace", request.Namespace)
+			return admission.Denied("deny evict mount pod because it's has juicefs reference")
```
zwwhdls:
```suggestion
			return admission.Denied("deny evict mount pod because it has juicefs reference")
```

zwwhdls (approved):

PR #1348, opened 2025-05-15
get csi  spec

showjason:
#1194 
To get CSI Spec from CSI Pod instead of CSI DaemonSet
Please review and comment

### DIFF
--- a/cmd/controller.go
+++ b/cmd/controller.go
@@ -25,6 +25,7 @@ import (
 
 	"github.com/prometheus/client_golang/prometheus/promhttp"
 	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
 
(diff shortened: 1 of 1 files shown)


zxh326 (changes requested):
<del>Maybe a better way is to use [Kubernetes' Downward API](https://kubernetes.io/zh-cn/docs/tasks/inject-data-application/environment-variable-expose-pod-information/) to pass the CSI pod's name/namespace to containers via environment variables, instead of using list.</del>

showjason:
> Maybe a better way is to use [Kubernetes' Downward API](https://kubernetes.io/zh-cn/docs/tasks/inject-data-application/environment-variable-expose-pod-information/) to pass the CSI pod's name/namespace to containers via environment variables, instead of using list.

Sorry, I'm a bit confused 🤣 ! Allow me to ask more and make it clear, in my understanding, this is the feature of `juicers-csi-controller`.
 [Kubernetes' Downward API](https://kubernetes.io/zh-cn/docs/tasks/inject-data-application/environment-variable-expose-pod-information/) can only support containers to get its own Pod information, but cannot get the external Pod(CSI node pod), hence I think it's infeasible to get CSI node pod by this way.

cmd/controller.go
```diff
@@ -100,17 +101,46 @@ func parseControllerConfig() {
...
 			os.Exit(1)
 		}
-		config.CSIPod = corev1.Pod{
-			Spec: ds.Spec.Template.Spec,
+		var csiPod *corev1.Pod
+		for i := range pods {
+			isCSINodePod := false
+			// Ensure the pod is managed by the expected DaemonSet
+			for _, ownerRef := range pods[i].OwnerReferences {
+				if ownerRef.Kind == "DaemonSet" && ownerRef.Name == CSINodeDsName {
```
zwwhdls: Do not need to check ownerRef here. Our goal is to support Advanced Daemonset. Check the length of list results. and use the first one.

zwwhdls (changes requested):

cmd/controller.go
```diff
@@ -100,17 +101,46 @@ func parseControllerConfig() {
...
 			CSINodeDsName = name
 		}
-		ds, err := k8sclient.GetDaemonSet(context.TODO(), CSINodeDsName, config.Namespace)
+		labelSelector := &metav1.LabelSelector{
+			MatchLabels: map[string]string{
+				"app": "juicefs-csi-node",
+			},
+		}
+
+		pods, err := k8sclient.ListPod(context.TODO(), config.Namespace, labelSelector, nil)
```
zxh326, line 111: add a new func, like ListPodWithLimit， and `listOptions.limit`=1?

cmd/controller.go
zxh326: Maybe it's not create by daemonset?

cc @zwwhdls

showjason: Yes, this is what I think.
With label `app=juicefs-csi-node` can not 100% guarantee the uniqueness of CSI node Pod,  and if any pod that is not part of juicefs-csi uses “app=juicefs-csi-node”, it can lead to incidents.
CC @zwwhdls

cmd/controller.go
showjason: 🤔 hmm, this action of getting the csi node pod only happens once when the csi controller starts, is it necessary to add a new func? 
However, perhaps there will be more scenarios where this method `ListPodWithLimit ` will be used in the future

cmd/controller.go
zwwhdls: Do not consider that corner case. We suppose just juicefs-csi-node use label "app=juicefs-csi-node", and we use this label also in dashboard to identify juicefs csi. Check the length of list result of labelSelector to guarantee it. 

As is discussed in https://github.com/juicedata/juicefs-csi-driver/issues/1194#issuecomment-2840677970

> Users usually do not modify this label. Even if it is modified, we believe that the configuration of CSI Node is incorrect, that is, they rely on this label.

People can only install juicefs csi by chart and k8s.yaml. The label "app=juicefs-csi-node" is both in them and **only** csi node has it.

zwwhdls: What's your opinion? @zxh326

zxh326: limit is set to 1, take the first pod, I think that is enough, no need to worry about special cases

showjason: Understood! Thanks for your patience and suggestions, let me update the code.

zxh326:
need rebase master to fix unit test cc @showjason

showjason:
> need rebase master to fix unit test cc @showjason

Will do that

zxh326 (approved):
LGTM

PR #1378, opened 2025-07-11
fix: check version in set quota with create flag

zxh326:
fix: https://github.com/juicedata/juicefs-csi-driver/actions/runs/16213030120/job/45776583069

### DIFF
--- a/pkg/driver/controller.go
+++ b/pkg/driver/controller.go
@@ -156,12 +156,14 @@ func (d *controllerService) CreateVolume(ctx context.Context, req *csi.CreateVol
 	}
 
 	if config.GlobalConfig.EnableControllerSetQuota == nil || *config.GlobalConfig.EnableControllerSetQuota {
-		volCtx[common.ControllerQuotaSetKey] = "true"
-		d.quotaPool.Run(context.Background(), func(ctx context.Context) {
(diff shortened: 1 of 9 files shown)


pkg/driver/controller.go
```diff
@@ -156,12 +156,14 @@ func (d *controllerService) CreateVolume(ctx context.Context, req *csi.CreateVol
 	}
 
 	if config.GlobalConfig.EnableControllerSetQuota == nil || *config.GlobalConfig.EnableControllerSetQuota {
-		volCtx[common.ControllerQuotaSetKey] = "true"
-		d.quotaPool.Run(context.Background(), func(ctx context.Context) {
-			if err := d.setQuotaInController(ctx, volumeId, req.GetCapacityRange(), options, subPath, secrets, volCtx); err != nil {
-				log.Error(err, "set quota in controller error")
-			}
-		})
+		if util.SupportQuotaPathCreate(true, config.BuiltinCeVersion) && util.SupportQuotaPathCreate(false, config.BuiltinEeVersion) {
```
zwwhdls, line 159: fallback to the old progress if it is not supported?

zxh326: If not supported, `common.ControllerQuotaSetKey` ctx will not set, and the setQuota will be handled in the node.

zwwhdls (approved):

PR #1394, opened 2025-08-12
fix: volume subpath is created with 755 permission

### LINKED ISSUE #1393: [BUG] volume subpath is created with 755 permission
**What happened**:

In v0.29.0, `enableControllerSetQuota` is set to true by default. This means `juicefs set quota --create` is executed on the juicefs-csi-controller.

This command creates a subdirectory within the volume. but, the permission might be 755 not 777.

As a result, the application can’t write to our juicefs unless root.

**Environment**:
- Others:

kyungwan-nam:
fix https://github.com/juicedata/juicefs-csi-driver/issues/1393

### DIFF
--- a/pkg/juicefs/juicefs.go
+++ b/pkg/juicefs/juicefs.go
@@ -661,7 +661,7 @@ func (j *juicefs) SetQuota(ctx context.Context, secrets map[string]string, jfsSe
 	}
 
 	cmdStr := fmt.Sprintf("umask 000; %s %s", config.CeCliPath, strings.Join(args, " "))
-	quotaCmd := j.Exec.CommandContext(ctx, "/bin/bash", "-c", cmdStr)
+	quotaCmd := j.Exec.CommandContext(ctx, "sh", "-c", cmdStr)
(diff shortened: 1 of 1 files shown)


pkg/juicefs/juicefs.go
```diff
@@ -649,7 +649,8 @@ func (j *juicefs) SetQuota(ctx context.Context, secrets map[string]string, jfsSe
 		if err != nil {
 			return errors.Wrap(err, authRes)
 		}
-		quotaCmd := j.Exec.CommandContext(cmdCtx, config.CliPath, args...)
+		cmdStr := fmt.Sprintf("umask 000; %s %s", config.CliPath, strings.Join(args, " "))
+		quotaCmd := j.Exec.CommandContext(cmdCtx, "/bin/bash", "-c", cmdStr)
```
zxh326:
```suggestion
		quotaCmd := j.Exec.CommandContext(cmdCtx, "sh", "-c", cmdStr)
```

zxh326 (commented):
LGTM

cc @zwwhdls

PR #1406, opened 2025-09-05
Setting quota should be configurable

### LINKED ISSUE #1405: [FEATURE] Setting quota should be configurable

Setting quota should be configurable

When a JuiceFS filesystem is shared by multiple users, the administrator wants to manually set quotas for each user and grant them a certain amount of storage.

However, since quotas are always set based on a user's PVC capacity, this conflicts with the quota set by the administrator.

I need a way to disable quota settings.

kyungwan-nam:
fix https://github.com/juicedata/juicefs-csi-driver/issues/1405
Added `EnableSetQuota` as a global config to disable quota settings.

### DIFF
--- a/pkg/driver/controller.go
+++ b/pkg/driver/controller.go
@@ -155,15 +155,19 @@ func (d *controllerService) CreateVolume(ctx context.Context, req *csi.CreateVol
 		}
 	}
 
-	if config.GlobalConfig.EnableControllerSetQuota == nil || *config.GlobalConfig.EnableControllerSetQuota {
-		if util.SupportQuotaPathCreate(true, config.BuiltinCeVersion) && util.SupportQuotaPathCreate(false, config.BuiltinEeVersion) {
-			volCtx[common.ControllerQuotaSetKey] = "true"
-			d.quotaPool.Run(context.Background(), func(ctx context.Context) {
-				if err := d.setQuotaInController(ctx, volumeId, req.GetCapacityRange(), options, subPath, secrets, volCtx); err != nil {
-					log.Error(err, "set quota in controller error")
-				}
-			})
+	if config.GlobalConfig.EnableSetQuota == nil || *config.GlobalConfig.EnableSetQuota {
+		if config.GlobalConfig.EnableControllerSetQuota == nil || *config.GlobalConfig.EnableControllerSetQuota {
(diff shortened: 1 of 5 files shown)


pkg/driver/controller.go
```diff
@@ -323,6 +327,9 @@ func (d *controllerService) ListSnapshots(ctx context.Context, req *csi.ListSnap
 
 // ControllerExpandVolume adjusts quota according to capacity settings
 func (d *controllerService) ControllerExpandVolume(ctx context.Context, req *csi.ControllerExpandVolumeRequest) (*csi.ControllerExpandVolumeResponse, error) {
+	if config.GlobalConfig.EnableSetQuota != nil && *config.GlobalConfig.EnableSetQuota == false {
```
zxh326:
```suggestion
	if config.GlobalConfig.EnableSetQuota != nil && !*config.GlobalConfig.EnableSetQuota {
```

pkg/driver/controller.go
```diff
@@ -155,15 +155,19 @@ func (d *controllerService) CreateVolume(ctx context.Context, req *csi.CreateVol
...
+				volCtx[common.ControllerQuotaSetKey] = "true"
+				d.quotaPool.Run(context.Background(), func(ctx context.Context) {
+					if err := d.setQuotaInController(ctx, volumeId, req.GetCapacityRange(), options, subPath, secrets, volCtx); err != nil {
+						log.Error(err, "set quota in controller error")
+					}
+				})
+			}
 		}
+	} else {
+		volCtx[common.DisableQuotaSetKey] = "true"
```
zxh326: once this ctx is added, it cannot be changed. If `EnableSetQuota` config changed to true, the quota cannot be set again.

I suggest not adding this ctx, and only checking the config in [node](https://github.com/juicedata/juicefs-csi-driver/pull/1406/files#diff-4f4d9c2e0ec5c4d3b4ac2d79abd70ee21bacff65ba05c059396921996dbf6607R198-R199).

zxh326 (approved):
LGTM cc @zwwhdls

zwwhdls (approved):

kyungwan-nam:
@zxh326 @zwwhdls 
Thanks for taking your time to review

PR #1426, opened 2025-10-22
fix snapshot

zxh326:
- compatible with ee
- support provisioner api to restore
- remove the e2e snapshot test because it's currently not being tested and requires installing an additional snapshot controller
- remove the snapshot deploy manifest, we need to provide a feature flag to enable this feature in helm repo
- remove listSnapshot api, not supported at the moment.

image

cc @DeanKamali  @deanmkk


DeanKamali:
@zxh326 I don't think I have permissions to review this one.

pkg/driver/provisioner.go
```diff
@@ -255,9 +266,70 @@ func (j *provisionerService) Provision(ctx context.Context, options provisioncon
...
+
+	snapshotId, sourceVolumeId, err := util.ParseSnapshotHandle(*snapContentObj.Status.SnapshotHandle)
+	if err != nil {
+		return fmt.Errorf("parse snapshot handle %s error: %v", *snapContentObj.Status.SnapshotHandle, err)
+	}
+	targetVolumeID := pv.Spec.CSI.VolumeHandle
+	targetSubPath := pv.Spec.CSI.VolumeAttributes["subPath"]
+	provisionerLog.Info("Restoring volume from snapshot", "snapshotId", snapshotId, "sourceVolumeId", sourceVolumeId, "targetVolumeID", targetVolumeID, "targetSubPath", targetSubPath)
+	secrets := make(map[string]string)
+	secretName := pv.Spec.CSI.NodePublishSecretRef.Name
```
zwwhdls, line 320: According to [csi document](https://kubernetes-csi.github.io/docs/secrets-and-credentials-volume-group-snapshot-class.html), maybe we should use snapshot-secret instead?

zxh326: snapshot-secret is set in VolumeGroupSnapshotClass and is used to create and delete volumeSnapshots. 

Here we use it for restore, using the secret from the target.

zwwhdls: In controllerService, controller-secret is used to restoreSnapshot which is in `CreateVolume`

zxh326: controllerService use provision secret

zxh326: https://github.com/kubernetes-csi/external-provisioner/blob/1b71152a5b3463d9c13157c488e54a00e7bedc9c/pkg/controller/controller.go#L711-L719

zwwhdls: But provision secret is not node-publish-secret? Maybe `pv.Spec.PersistentVolumeSource.CSI.VolumeAttributes[common.ProvisionerSecretName]` should be used here.

zwwhdls (approved):

PR #1430, opened 2025-10-27
fix: use HostPathDirectoryOrCreate for clean cache pod volumes

### LINKED ISSUE #1429: [BUG] Clean cache pod fails when raw cache directory doesn't exist on first use
**What happened**:
When a mount pod terminates and triggers clean cache pod creation, the pod fails to start with `ContainerStatusUnknown` status. Kubelet logs show:
```
Error: MountVolume.SetUp failed for volume "cachedir-0" (UniqueName: "kubernetes.io/host-path/770d74f3-f62f-4e6d-b9f7-da5accebcf8b-cachedir-0") pod "juicefs-4ae9ea24-cleancache-idnsuf-b6kdp" (UID: "770d74f3-f62f-4e6d-b9f7-da5accebcf8b") : hostPath type check failed: /var/jfsCache/332eecd3-7c33-49f7-bf4d-a287d32dc192/raw is not a directory
```

Clean cache pod should start successfully even when the `raw` subdirectory doesn't exist.

1. Use JuiceFS CSI Driver on a fresh node where JuiceFS volume has never been used
2. Create a pod that uses JuiceFS volume with cache-dir configured
3. Delete the pod immediately **before any cache operations occur** (before the `raw` directory is created by actual cache usage)
4. Mount pod will be automatically deleted, then clean cache pod is created
5. Observe clean cache pod fails with hostPath mount error

Since the `raw` directory doesn't exist, the clean cache pod failure doesn't affect the actual cache cleaning operation (there's no cache to clean anyway). However, this creates confusion for users with error messages and abnormal pod status.

**Environment**:
- JuiceFS CSI Driver version (which image tag did your CSI Driver use): latest
- Kubernetes version (e.g. `kubectl version`): v1.34.0
- Others:

Looka149:
Change `HostPathDirectory` to `HostPathDirectoryOrCreate` in `genCleanCachePod()` to allow automatic directory creation if it doesn't exist.

Resolves https://github.com/juicedata/juicefs-csi-driver/issues/1429

### DIFF
--- a/pkg/juicefs/mount/builder/pod.go
+++ b/pkg/juicefs/mount/builder/pod.go
@@ -302,7 +302,7 @@ func (r *PodBuilder) genCleanCachePod() *corev1.Pod {
 	cacheVolumes := []corev1.Volume{}
 	cacheVolumeMounts := []corev1.VolumeMount{}
 
-	hostPathType := corev1.HostPathDirectory
+	hostPathType := corev1.HostPathDirectoryOrCreate
 
 	for idx, cacheDir := range r.jfsSetting.CacheDirs {
 		name := fmt.Sprintf("cachedir-%d", idx)


zxh326 (approved):

Looka149:
@zxh326 Thank you for the review and merge :)

PR #1431, opened 2025-10-28
feat(sidecar): attempts to use req.namespace

nioshield:
In some scenarios, the pod's namespace may be empty; you can try using req.Namespace.


zxh326 (approved):
LGTM

cc @zwwhdls

zwwhdls (approved):

PR #1434, opened 2025-11-04
fix: delete volumePath metrics when unpublish

zxh326:
1. `NodeUnpublishVolume` successfully unmounts the mount path
2. The app Pod exits
3. Kubelet calls `NodeGetVolumeStats` to check the volume status
4. The mount point has already been unmounted, the `volumePathHealth` metric is set to 0 (not mounted)
5. Since the Pod has already exited, Kubelet won't call again, and the metric remains at 0 forever

### DIFF
--- a/pkg/driver/node.go
+++ b/pkg/driver/node.go
@@ -345,8 +345,8 @@ func (d *nodeService) NodeGetVolumeStats(ctx context.Context, req *csi.NodeGetVo
 	}
 
 	if d.isPathUnmounted(volumePath) {
-		log.Info("Volume path was unmounted, ignoring stats request", "volumePath", volumePath)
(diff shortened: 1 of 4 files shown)


pkg/driver/node.go
```diff
@@ -309,6 +344,11 @@ func (d *nodeService) NodeGetVolumeStats(ctx context.Context, req *csi.NodeGetVo
 		return nil, status.Error(codes.InvalidArgument, "Volume path not provided")
 	}
 
+	if d.isPathUnmounted(volumePath) {
+		log.Info("Volume path was unmounted, ignoring stats request", "volumePath", volumePath)
```
zwwhdls: More details here to distinguish the situation with "not mounted error"

zwwhdls (approved):

PR #1460, opened 2025-12-17
fix: use podname as hostname when hostnamekey is set to 'podname' in hostNetwork

zxh326:


zxh326:
not work, issue  https://github.com/kubernetes/kubernetes/issues/67019

PR #1463, opened 2025-12-19
support disable mount hostpath

kungf:
The mounting of the hostPath volume (/var/jfsCache) is conditional， it can be disabled if the default cache directory is selected, in which case the app pod's default data disk is used. The hostPath volume is mounted only when a specific cache-dir is explicitly defined.

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -199,9 +199,10 @@ type MountPodPatch struct {
 	// omit will patch for all PVC
 	PVCSelector *PVCSelector `json:"pvcSelector,omitempty"`
 
-	CEMountImage string               `json:"ceMountImage,omitempty"`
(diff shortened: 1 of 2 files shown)


zxh326 (changes requested):
This will result in all mountpods being non-reusable after the upgrade csi-node.

zxh326:
Have you tried using emptyDir?

kungf:
> This will result in all mountpods being non-reusable after the upgrade csi-node.

"DisableCacheHostPath" defaults to false. Upgrades do not affect the existing logic unless "DisableCacheHostPath" is explicitly set to true. This indicates a clear intention not to use the default hostpath mount, so it has no impact on the upgrade.

kungf:
> Have you tried using emptyDir?

it seems not work

PR #1464, opened 2025-12-19
fix: make fuse abort job script POSIX-safe

kungf:
handle missing sysfs file err
```
+ supFusePass=false
+ [ false = true ]
+ cat /sys/fs/fuse/connections/803/waiting
cat: /sys/fs/fuse/connections/803/waiting: No such file or directory
+ [ -eq 0 ]
sh: 15: [: -eq: unexpected operator
+ echo fuse mount point is hung or deadlocked, aborting...
sh: 20: cannot create /sys/fs/fuse/connections/803/abort: Directory nonexistent
+ echo 1
fuse mount point is hung or deadlocked, aborting...
```

### DIFF
--- a/pkg/juicefs/mount/builder/job.go
+++ b/pkg/juicefs/mount/builder/job.go
@@ -182,9 +182,9 @@ func NewFuseAbortJob(mountpod *corev1.Pod, devMinor uint32, mntPath string) *bat
 	supFusePass := util.SupportFusePass(mountpod)
 	command := fmt.Sprintf(`set -x
 supFusePass=%t
-if [ $supFusePass = true ]; then
+if [ "$supFusePass" = "true" ]; then
   attempt=1
-  while [ $attempt -le 5 ]; do
+	while [ "$attempt" -le 5 ]; do
     if inode=$(timeout 1 stat -c %%i %s 2>/dev/null) && [ "$inode" = "1" ]; then
       echo "fuse mount point is normal, exit 0"
       exit 0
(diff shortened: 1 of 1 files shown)


pkg/juicefs/mount/builder/job.go
```diff
@@ -182,9 +182,9 @@ func NewFuseAbortJob(mountpod *corev1.Pod, devMinor uint32, mntPath string) *bat
 	supFusePass := util.SupportFusePass(mountpod)
 	command := fmt.Sprintf(`set -x
 supFusePass=%t
-if [ $supFusePass = true ]; then
+if [ "$supFusePass" = "true" ]; then
   attempt=1
-  while [ $attempt -le 5 ]; do
+	while [ "$attempt" -le 5 ]; do
```
zxh326:
```suggestion
  while [ "$attempt" -le 5 ]; do
```

pkg/juicefs/mount/builder/job.go
```diff
@@ -194,13 +194,35 @@ if [ $supFusePass = true ]; then
...
+
+# Some kernels may not provide waiting/abort files, handle gracefully.
+if [ -f "$waiting_file" ]; then
+	waiting=$(cat "$waiting_file" 2>/dev/null || echo "")
+	if [ -n "$waiting" ] && [ "$waiting" -eq 0 ] 2>/dev/null; then
+		echo "fuse connections 'waiting' is zero, skip"
+		exit 0
+	fi
+else
+	echo "fuse connections 'waiting' file not found: $waiting_file, continue"
```
zxh326: skip here? 

It's possible that the mountpoint has already umount normally, so it does not exist.

pkg/juicefs/mount/builder/job.go
```diff
@@ -194,13 +194,35 @@ if [ $supFusePass = true ]; then
...
 fi
 
 echo "fuse mount point is hung or deadlocked, aborting..."
-echo 1 > /sys/fs/fuse/connections/%d/abort
-`, supFusePass, mntPath, devMinor, devMinor)
+if [ -w "$abort_file" ]; then
+	echo 1 > "$abort_file"
+else
+	echo "fuse connections 'abort' file not writable/not found: $abort_file"
```
zwwhdls:
```suggestion
	echo "fuse connections 'abort' file not writable/found: $abort_file"
```

pkg/juicefs/mount/builder/job.go
kungf: fix

pkg/juicefs/mount/builder/job.go
kungf: fix

pkg/juicefs/mount/builder/job.go
kungf: fix

pkg/juicefs/mount/builder/job.go
zxh326: When the `waiting` equals 0 or the `waiting` file doesn't exist, it's better to skip the process and exit with code 0.

zxh326 (commented):
Can you convert all the tabs in the script to 4 spaces?

pkg/juicefs/mount/builder/job.go
kungf: fix

zxh326 (approved):

PR #1468, opened 2025-12-22
mount pod support set tolerations

kungf:


zxh326:
Mountpod defaults to inheriting the tolerations from the business pod. Is there still a need for this?

kungf:
> Mountpod defaults to inheriting the tolerations from the business pod. Is there still a need for this?

This feature does not affect the old inheritance logic, supports custom tolerations, and is more robust.

zxh326:
> > Mountpod defaults to inheriting the tolerations from the business pod. Is there still a need for this?
> 
> This feature does not affect the old inheritance logic, supports custom tolerations, and is more robust.

My mean is why this feature is still needed. The mountpod must run on the node where the business is running. If the business pod runs successfully, it can be considered to have already included appropriate tolerations.

In addition, mountpod uses nodename by default to bypass the scheduler.

PR #1471, opened 2025-12-24
fix: support cleanup of pending mount pods with custom schedulers

kungf:
This commit addresses the issue where mount pods with custom schedulers
remain stuck in Pending state forever and are not cleaned up when their
corresponding app pods are deleted.

Root Cause:
When mount pods use custom schedulers, they may remain in Pending state
(spec.nodeName is empty) if the scheduler rejects them.  The csi not will not
delete it.

Testing:
Verified that pending mount pods are correctly cleaned up when their
app pods are deleted, even when using custom schedulers.

### DIFF
--- a/pkg/controller/mount_controller.go
+++ b/pkg/controller/mount_controller.go
@@ -64,9 +64,18 @@ func (m MountController) Reconcile(ctx context.Context, request reconcile.Reques
 		return reconcile.Result{}, nil
 	}
 
+	// Scenario 1: Handle pending mount pod without nodeName (not scheduled yet)
+	// This handles the case where mount pod is stuck in Pending state due to custom scheduler
+	if mountPod.DeletionTimestamp == nil &&
+		mountPod.Status.Phase == corev1.PodPending &&
+		mountPod.Spec.NodeName == "" {
+		return m.handlePendingMountPod(ctx, mountPod)
+	}
+
+	// Scenario 2: Handle mount pod being deleted (original logic)
 	// check mount pod deleted
 	if mountPod.DeletionTimestamp == nil {
-		mountCtrlLog.V(1).Info("pod is not deleted", "name", mountPod.Name)
+		mountCtrlLog.V(1).Info("pod is not deleted and not pending", "name", mountPod.Name)
 		return reconcile.Result{}, nil
 	}
 	if !util.ContainsString(mountPod.GetFinalizers(), common.Finalizer) {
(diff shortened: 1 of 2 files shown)


kungf:
@zxh326 update

pkg/controller/mount_controller.go
```diff
@@ -102,6 +111,126 @@ func (m MountController) Reconcile(ctx context.Context, request reconcile.Reques
...
+					"appPodUid", targetUid,
+					"mountPod", mountPod.Name)
+			}
+		}
+	}
+
+	// If no app pods reference this mount pod anymore, delete it
+	if existingRefs == 0 {
+		// Check if should delay deletion
+		shouldDelay, err := resource.ShouldDelay(ctx, mountPod, m.K8sClient)
```
zxh326: pending pod does not need to consider delay

pkg/controller/mount_controller.go
```diff
@@ -102,6 +111,126 @@ func (m MountController) Reconcile(ctx context.Context, request reconcile.Reques
...
+
+	var pods *corev1.PodList
+	var err error
+
+	if nodeName != "" {
+		// Optimized: Only list pods on the target node
+		fieldSelector := fields.Set{"spec.nodeName": nodeName}.AsSelector().String()
+		pods, err = m.K8sClient.CoreV1().Pods("").List(ctx, metav1.ListOptions{
+			FieldSelector: fieldSelector,
+		})
```
zxh326:
```suggestion
		labelSelector := metav1.LabelSelector{
			MatchExpressions: []metav1.LabelSelectorRequirement{{
				Key:      common.UniqueId,
				Operator: metav1.LabelSelectorOpExists,
			}},
		}
		labelSelectorStr, _ := metav1.LabelSelectorAsSelector(&labelSelector)
		fieldSelector := fields.Set{"spec.nodeName": nodeName}.AsSelector().String()
		pods, err = m.K8sClient.CoreV1().Pods("").List(ctx, metav1.ListOptions{
			FieldSelector: fieldSelector,
			LabelSelector: labelSelectorStr.String(),
		})
```

kungf: update

pkg/controller/mount_controller.go
kungf: delete

pkg/controller/mount_controller.go
```diff
@@ -102,6 +111,124 @@ func (m MountController) Reconcile(ctx context.Context, request reconcile.Reques
...
+
+	// If no app pods reference this mount pod anymore, delete it
+	if existingRefs == 0 {
+		mountCtrlLog.Info("No app pods reference this pending mount pod, deleting it",
+			"mountPod", mountPod.Name,
+			"namespace", mountPod.Namespace)
+
+		if err := m.DeletePod(ctx, mountPod); err != nil {
+			mountCtrlLog.Error(err, "Failed to delete pending mount pod", "mountPod", mountPod.Name)
+			return reconcile.Result{}, err
```
zxh326:
```suggestion
			return reconcile.Result{}, client.IgnoreNotFound(err)
```

pkg/controller/mountinfo.go
```diff
@@ -230,6 +230,7 @@ type mountItem struct {
 }
 
 func getPodUid(target string) string {
+	// Target format: /var/lib/kubelet/pods/<pod-uid>/volumes/kubernetes.io~csi/<volume-id>/mount
```
zxh326, line 233: keep

zxh326 (approved):
/LGTM cc @zwwhdls

zwwhdls (approved):

PR #1478, opened 2026-01-12
fix: remove incorrect shell escaping for env vars passed to exec

### LINKED ISSUE #1477: [BUG] special characters in metadata engine secret cause incorrect behavior
**What happened**:

Using Redis cluster with password as metadata engine.
When password contains special characters like `admin$`, mount fails with incorrect behavior.

It seems `security.EscapeBashStr` causes unintended escaping in this case.
https://github.com/juicedata/juicefs-csi-driver/blob/ebdc67230bc63f6c208ef05293d6b98606cca2d5/pkg/juicefs/juicefs.go#L903-L905

**Environment**:
- Others:

Looka149:
Resolves https://github.com/juicedata/juicefs-csi-driver/issues/1477

### DIFF
--- a/pkg/config/command.go
+++ b/pkg/config/command.go
@@ -207,7 +207,7 @@ func GetJfsVolUUID(ctx context.Context, s *JfsSetting) (string, error) {
 	statusCmd := exec.CommandContext(cmdCtx, CeCliPath, "status", s.Source)
 	envs := syscall.Environ()
 	for key, val := range s.Envs {
-		envs = append(envs, fmt.Sprintf("%s=%s", security.EscapeBashStr(key), security.EscapeBashStr(val)))
+		envs = append(envs, fmt.Sprintf("%s=%s", key, val))
 	}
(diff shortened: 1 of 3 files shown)


zxh326 (commented):
setEnv `escapeBashStr` is indeed not needed.，not parsed by bash/shell

/LGTM

cc @zwwhdls

zxh326:
Can you also remove this from other functions, such as setQuota/format?

Looka149:
@zxh326 Thank you for your review.

> Can you also remove this from other functions, such as setQuota/format?

This has already been applied to the `SetQuota` and `ceFormat` functions. Have I missed anything else?

Thanks

zwwhdls:
Good, LGTM. Can you please fix the test?

zxh326:
should be a flaky test. ignore

zxh326 (approved):

Looka149:
@zxh326 @zwwhdls 
Thank you for the review and merge.

PR #1479, opened 2026-01-15
fix: force update initconfig when secret fields change

zxh326:


pkg/controller/secret_controller.go
```diff
@@ -152,7 +157,27 @@ func refreshSecretInitConfig(ctx context.Context, client *k8sclient.K8sClient, n
 		return err
 	}
 	confs := string(b)
-	if _, ok := secretsMap["initconfig"]; ok {
+	hashSecretsMap := make(map[string]string)
+	maps.Copy(hashSecretsMap, secretsMap)
+	config.KeysCompatible(hashSecretsMap)
+	hashFields := []string{"token", "name", "access-key", "secret-key", "bucket", "envs"}
```
zwwhdls: add access-key2 and secret-key2

zwwhdls (approved):

PR #1591, opened 2026-05-29
feat: add juicefs_volume_info metric

### LINKED ISSUE #1590: [FEATURE] Expose pod and PVC identity as juicefs_volume_info metric

Expose the namespace, pod name and PVC name of each mount as a new metric `juicefs_volume_info`.

`juicefs_volume_path_health` only has `volume_id`, `volume_path` and `pod_uid` labels. When a mount is unhealthy, an operator cannot tell which namespace / pod / PVC it belongs to without an extra lookup. Joining with `kube_pod_info` / `kube_persistentvolumeclaim_info` is the usual way, but that requires kube-state-metrics, which is not always available.

---

**Proposed metric**

```
juicefs_volume_info{volume_id, pod_uid, pod_namespace, pod_name, pvc_name} 1
```

The value is always `1`; the identity is in the labels, as in `kube_pod_info`. `pod_namespace` is used instead of `namespace` to avoid colliding with the scrape target's own `namespace` label.

Join with the health metric on the shared `(volume_id, pod_uid)`:

```promql
juicefs_volume_path_health
  * on(volume_id, pod_uid) group_left(pod_namespace, pod_name, pvc_name)
    juicefs_volume_info
```

**How it works**

- `NodePublishVolume` sets the series. The PV is already fetched there, so no extra API call.
- `NodeUnpublishVolume` removes the series.
- On csi-node restart, kubelet does not re-publish already-mounted volumes, so a one-time bootstrap on node start repopulates the metric for them.

**API cost**

- Normal mount / unmount: no extra API call.
- No periodic calls.
- Only on csi-node restart: one node-scoped pod LIST, plus a PVC GET and PV GET per PVC referenced by pods on that node (not only JuiceFS ones).

**Known limitation**

If bootstrap fails at startup (e.g. the API server is unreachable), existing mounts have no `volume_info` series until the next csi-node restart or until the workload pod is recreated. The health metric is not affected.

Looka149:
Resolves #1590

Add a new metric `juicefs_volume_info{volume_id, pod_uid, pod_namespace, pod_name, pvc_name}` so `juicefs_volume_path_health` can be joined on `(volume_id, pod_uid)` to know the namespace / pod / PVC of a mount without kube-state-metrics.

The series is set in `NodePublishVolume`, removed in `NodeUnpublishVolume`, and repopulated once at csi-node startup for mounts that remain after a restart.

### DIFF
--- a/cmd/node.go
+++ b/cmd/node.go
@@ -221,6 +221,9 @@ func nodeRun(ctx context.Context) {
 		os.Exit(1)
 	}
 
+	// Node-binary only: the controller also builds a nodeService via NewDriver.
+	go drv.BootstrapVolumeInfo()
+
 	go func() {
 		<-ctx.Done()
 		drv.Stop()
--- a/pkg/driver/node.go
+++ b/pkg/driver/node.go
@@ -31,6 +31,9 @@ import (
 	"github.com/container-storage-interface/spec/lib/go/csi"
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/fields"
+	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/util/retry"
 	"k8s.io/klog/v2"
 	k8sexec "k8s.io/utils/exec"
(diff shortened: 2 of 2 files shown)


zxh326 (commented):
```
	metrics.volumePathHealth = prometheus.NewGaugeVec(prometheus.GaugeOpts{
		Name: "volume_path_health",
		Help: "health status of volume path (1 = healthy, 0 = unhealthy)",
	}, []string{"volume_id", "volume_path", "pod_uid"})
```


Is the `volume_path_health` information sufficient?

Looka149:
@zxh326 Thanks for the review.

With kube-state-metrics, joining `volume_path_health` on `volume_id`/`pod_uid` gives the namespace/pod/PVC. 
Without it, that information isn't available, so this metric provides it.

I made it a separate metric instead of adding labels to `volume_path_health` because health is set in `NodeGetVolumeStats`, whose CSI request has no pod/PVC information. `NodePublishVolume` already has that information (PodInfoOnMount + PV claimRef), so it can set the labels without an extra API call. The existing `volume_path_health` labels stay unchanged.

Looka149:
@zxh326 friendly ping.

This isn't specific to our environment. Any cluster without kube-state-metrics cannot tell which namespace/pod/PVC a mount belongs to.

It also helps debugging: when a mount is unhealthy, joining this with `juicefs_volume_path_health` shows which user pod is affected.

If you don't want it on by default, I can put it behind a flag (off by default), so it has no impact unless enabled. Would that work?

Looka149:
image

Using this metric, we can track app pods and mount pods from namespace and PVC information, as shown in the picture, without any additional installation.

zxh326:
> image
> Using this metric, we can track app pods and mount pods from namespace and PVC information, as shown in the picture, without any additional installation.

OK, I will consider this matter soon.

cmd/node.go
```diff
@@ -221,6 +221,9 @@ func nodeRun(ctx context.Context) {
 		os.Exit(1)
 	}
 
+	// Node-binary only: the controller also builds a nodeService via NewDriver.
+	go drv.BootstrapVolumeInfo()
```
zxh326, line 225: I suggest removing BootstrapVolumeInfo. Normally, NodePublishVolume should populate volume_info. After a csi-node restart, if the corresponding information is missing in NodeGetVolumeStats, extract the Pod UID and PV name from volumePath, obtain the PVC from the PV ClaimRef, find the Pod by UID from a node-scoped Pod LIST, and then populate both volume_info and the in-memory mapping.

Looka149: @zxh326 Sorry for the late reply, and thanks for the review. Done in 8051f77.

- `BootstrapVolumeInfo` is gone, so `cmd/node.go` is untouched now.
- `NodeGetVolumeStats` populates `volume_info` when it is missing.
- The in-memory mapping tracks which mounts already have `volume_info`, so repeated stats calls make no API call.

pkg/driver/node.go
```diff
@@ -445,3 +482,48 @@ func extractPodUIDFromVolumePath(volumePath string) string {
...
+	}
+	pv, err := d.k8sClient.GetPersistentVolume(ctx, pvName)
+	if err != nil {
+		log.V(1).Info("get PV failed", "pv", pvName, "error", err)
+		return
+	}
+	if pv.Spec.ClaimRef == nil || pv.Spec.ClaimRef.Name == "" {
+		return
+	}
+	pods, err := d.k8sClient.ListPod(ctx, pv.Spec.ClaimRef.Namespace, nil,
```
zxh326, line 513: After a csi-node restart/upgrade, each (volumeID, podUID) cache miss triggers a PV GET and Pod LIST, even when pods share the same volume. Could we reuse these lookup results across mounts to avoid repeated API requests during upgrades?

PR #1630, opened 2026-07-24
fix: parse image tag correctly when registry has a port

xiaojunxiang2023:
## Summary

- Fix `parseClientVersionFromImage` so image tags are taken from the last path component instead of splitting the whole reference on the first `:`.
- Private registries that expose a non-default port (e.g. `hub.example.com:8443/project/mount:ce-v1.2.3`) were previously misparsed: the "tag" became `8443/project/mount:ce-v1.2.3`, `SupportFusePass` returned false, and Mount Pods incorrectly got a default `preStop` `umount` hook. That breaks smooth upgrade / mount-point recovery for CSI >= 0.25.
- Also strip an image digest (`@sha256:...`) before parsing the tag.

## Test plan

- [x] `go test -gcflags=all=-l ./pkg/util/ -run 'TestParseClientVersion|TestSupportFusePassPod'` — passed (covers existing cases plus registry-port / digest cases)
- [x] Online validation before the code fix: using a mount image reference without a registry port made CSI stop injecting `preStop umount` for `ce-v1.2.3`; with `:8443` in the reference it kept injecting

### DIFF
--- a/pkg/util/util.go
+++ b/pkg/util/util.go
@@ -584,12 +584,23 @@ func parseClientVersionFromImage(image string) ClientVersion {
 	if image == "" {
 		return ClientVersion{}
 	}
-	imageSplits := strings.SplitN(image, ":", 2)
+	// Strip digest so "repo:tag@sha256:..." / "repo@sha256:..." are handled correctly.
+	if at := strings.LastIndex(image, "@"); at >= 0 {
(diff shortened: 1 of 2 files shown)


zxh326 (approved):
LGTM 

/cc @zwwhdls

PR #1633, opened 2026-07-28
allow upgrade dev image

zwwhdls:
1. allow upgrade dev image
2. ignore mountPath when diff config in dashboard

### DIFF
--- a/pkg/util/util.go
+++ b/pkg/util/util.go
@@ -690,7 +690,6 @@ func SupportFusePass(pod *corev1.Pod) bool {
 
 	image := pod.Spec.Containers[0].Image
 	v := parseClientVersionFromImage(image)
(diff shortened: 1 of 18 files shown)


pkg/util/util.go
```diff
@@ -690,11 +690,12 @@ func SupportFusePass(pod *corev1.Pod) bool {
 
 	image := pod.Spec.Containers[0].Image
 	v := parseClientVersionFromImage(image)
+	utilLog.Info("image", "nightly", v.Nightly, "dev", v.Dev)
```
zxh326: 改成 debug？

zwwhdls: 本地 debug 用的，可以删了

zxh326 (approved):

PR #1639, opened 2026-08-01
fix: make kubelet access error counting race-free and per-client

### LINKED ISSUE #1638: [BUG] data race on the global kubelet access error counter
**What happened**:

`kubeletAccessErrCount` is a package-level plain `int`, updated inside `GetNodeRunningPods` and calling `os.Exit(1)` after 5 consecutive failures.
Since #1505 it is updated from both the reconciler goroutine and concurrent mount handlers, which causes two problems:

1. The unsynchronized `++` / `= 0` is a data race: increments and resets can be lost.
2. The two independent clients mix their failure streaks in one global counter: a success from one client resets the other's streak.

https://github.com/juicedata/juicefs-csi-driver/blob/9c62c73bc15c33a1e28c503373a9753c34ddb712/pkg/k8sclient/kubelet_client.go#L179-L189

Each client counts its own consecutive failures, race-free, so the process exits only after 5 genuine consecutive failures of a single client.

Add this test to `pkg/k8sclient` and run `go test -race -run TestKubeletAccessErrCountRace ./pkg/k8sclient/`:

```go
func TestKubeletAccessErrCountRace(t *testing.T) {
	var wg sync.WaitGroup
	for i := 0; i < 2; i++ {
		wg.Add(1)
		go func() {
			defer wg.Done()
			for j := 0; j < 1000; j++ {
				checkKubeletAccessErr(errors.New("kubelet unreachable"))
				checkKubeletAccessErr(nil)
			}
		}()
	}
	wg.Wait()
}
```

```
WARNING: DATA RACE
Read at 0x000104cb4398 by goroutine 24:
  github.com/juicedata/juicefs-csi-driver/pkg/k8sclient.checkKubeletAccessErr()
      pkg/k8sclient/kubelet_client.go:184 +0x3c
Previous write at 0x000104cb4398 by goroutine 23:
  github.com/juicedata/juicefs-csi-driver/pkg/k8sclient.checkKubeletAccessErr()
      pkg/k8sclient/kubelet_client.go:184 +0x54
...
--- FAIL: TestKubeletAccessErrCountRace (0.00s)
    testing.go:1617: race detected during execution of test
```

I'll submit a PR that fixes this, keeping the current exit policy unchanged.
As a possible follow-up: mount-path callers already fall back to the api-server on kubelet errors, so their failures may not need to count toward the process exit. I can open a separate PR for that if you agree.

**Environment**:
- JuiceFS CSI Driver version (which image tag did your CSI Driver use): master (9c62c73)
- Others:

Looka149:
Resolves #1638

### DIFF
--- a/pkg/k8sclient/kubelet_client.go
+++ b/pkg/k8sclient/kubelet_client.go
@@ -23,6 +23,7 @@ import (
 	"net/http"
 	"os"
 	"strconv"
+	"sync/atomic"
 	"time"
 
 	corev1 "k8s.io/api/core/v1"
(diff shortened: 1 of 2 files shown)


zxh326 (approved):

Looka149:
@zxh326 Thank you for the review and merge.

PR #1643, opened 2026-08-05
precheck before batch upgrade

zwwhdls:
1. skip pod upgrade when node is SchedulingDisabled, which means the node may be something wrong.
2. skip pod upgrade when csi node is not ready
3. make timeout of batch pod upgrade configurable

### DIFF
--- a/cmd/dashboard/upgrade.go
+++ b/cmd/dashboard/upgrade.go
@@ -47,6 +47,9 @@ import (
 
 var batchConfigName string
 
+const batchUpgradeTimeoutEnv = "BATCH_UPGRADE_TIMEOUT"
+const defaultPodUpgradeTimeout = 300 * time.Second
+
 var upgradeCmd = &cobra.Command{
 	Use:   "upgrade",
 	Short: "trigger upgrade mount pod smoothly",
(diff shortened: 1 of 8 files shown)


cmd/dashboard/upgrade.go
```diff
@@ -557,6 +619,22 @@ func (u *BatchUpgrade) waitForUpgrade(ctx context.Context, index int, nodeName,
 	}
 }
 
+func getBatchUpgradeTimeout() (time.Duration, error) {
+	value := os.Getenv(batchUpgradeTimeoutEnv)
+	if value == "" {
+		return defaultPodUpgradeTimeout, nil
+	}
+	seconds, err := strconv.ParseInt(value, 10, 64)
```
zxh326, line 627: 看你的单测是支持 45s 这类写法

zxh326 (approved):

cmd/dashboard/upgrade.go
zwwhdls: 修复了单测，同时在 dashboard 里新建 Job 前检查环境变量，只允许是整数，表示秒

PR #1662, opened 2026-08-25
fix: accept the byte suffix in buffer-size

### LINKED ISSUE #1661: [BUG] buffer-size with a byte suffix is silently dropped
**What happened**:

JuiceFS takes `buffer-size=1073741824B`, but the CSI driver drops it.

```
E0825 09:10:41.682147      13 setting.go:1200] "parse buffer-size error, ignore buffer-size option" err="cannot parse 1073741824B to bytes, invalid unit" logger="config" buffer-size="1073741824B"
```

```yaml
annotations:
  csi.juicefs.com/additional-mount-options: "buffer-size=1073741824B"
```

A mount pod memory limit must be set, otherwise the value is never parsed.

I'll submit a PR.

**Environment**:
- JuiceFS CSI Driver version (which image tag did your CSI Driver use): v0.32.4. Same code on master (1aa7e43)
- Kubernetes version (e.g. `kubectl version`): v1.34.0 (kind)
- Others: mount image `juicedata/mount:ce-v1.4.1`

Looka149:
Resolves #1661

### DIFF
--- a/pkg/util/util.go
+++ b/pkg/util/util.go
@@ -513,6 +513,7 @@ func ParseToBytes(value string) (uint64, error) {
 	}
 	var shift int
 	switch unit {
+	case 'B':
 	case 'k', 'K':
 		shift = 10
 	case 'm', 'M':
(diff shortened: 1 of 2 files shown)


zxh326 (approved):

Looka149:
@zxh326 Thank you for the review and merge.

PR #1664, opened 2026-08-25
fix: drop a buffer-size percentage that has no memory limit to size it

### LINKED ISSUE #1663: [BUG] Mount pod crash loops when buffer-size uses a percentage without a memory limit
**What happened**:

A percentage `buffer-size` is not converted when the mount pod has no memory limit. The raw string reaches the mount process, which exits before mounting.

```
juicefs[1] <INFO>: Ping redis latency: 54.478µs
juicefs[1] <FATAL>: Invalid value "30%" for "buffer-size": invalid unit [ParseBytesStr@humanize.go:63]
```

`processOption` returns the option unchanged without a memory limit, and `genAndValidOptions` skips its own check on the same condition, so nothing catches it.

https://github.com/juicedata/juicefs-csi-driver/blob/1aa7e43a43bc0fe7d467743667f2577e51a5eb6f/pkg/config/setting.go#L1166-L1168
https://github.com/juicedata/juicefs-csi-driver/blob/1aa7e43a43bc0fe7d467743667f2577e51a5eb6f/pkg/config/setting.go#L576-L579

The mount pod ends up in CrashLoopBackOff. The PVC stays `Bound`, so only the application pod shows it, stuck in `ContainerCreating` behind a 60s mount timeout per retry.

The mount should not fail.

Set both annotations on a PVC. Both are documented: [percentage `buffer-size`](https://juicefs.com/docs/csi/guide/resource-optimization/#sidecar-resource-percentages) and [`"0"` to omit a limit](https://juicefs.com/docs/csi/guide/resource-optimization/#omit-resources). A patch that sets only `resources.requests` reaches the same state, since `attr.Resources` is replaced whole.

```yaml
annotations:
  juicefs/mount-memory-limit: "0"
  csi.juicefs.com/additional-mount-options: "buffer-size=30%"
```

**Environment**:
- JuiceFS CSI Driver version (which image tag did your CSI Driver use): v0.32.4, chart 0.32.4. Same code on master (1aa7e43)
- Kubernetes version (e.g. `kubectl version`): v1.34.0 (kind)
- Others: mount image `juicedata/mount:ce-v1.4.1`, Redis metadata engine, MinIO object storage

Looka149:
Resolves #1663

### DIFF
--- a/pkg/config/setting.go
+++ b/pkg/config/setting.go
@@ -1156,14 +1156,20 @@ func getDefaultResource() corev1.ResourceRequirements {
 	}
 }
 
-func processOption(option string, resources corev1.ResourceRequirements) string {
+func processOption(option string, resources corev1.ResourceRequirements, finalPass bool) string {
 	pair := strings.Split(option, "=")
-	if len(pair) != 2 || pair[0] != "buffer-size" {
+	if len(pair) != 2 || strings.TrimSpace(pair[0]) != "buffer-size" {
 		return option
 	}
+	pair[0], pair[1] = strings.TrimSpace(pair[0]), strings.TrimSpace(pair[1])
 	memLimit := resources.Limits[corev1.ResourceMemory]
 	memLimitByte := memLimit.Value()
 	if memLimitByte <= 0 {
+		// sidecar resources are only settled in the final pass, so until then a percentage may still be waiting for its limit
+		if finalPass && strings.HasSuffix(pair[1], "%") {
+			log.Info("dropping buffer-size percentage without a memory limit, JuiceFS falls back to its default", "option", option)
+			return ""
+		}
 		return option
 	}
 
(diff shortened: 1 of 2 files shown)


pkg/config/setting.go
```diff
@@ -1156,14 +1156,20 @@ func getDefaultResource() corev1.ResourceRequirements {
...
-	if len(pair) != 2 || pair[0] != "buffer-size" {
+	if len(pair) != 2 || strings.TrimSpace(pair[0]) != "buffer-size" {
 		return option
 	}
+	pair[0], pair[1] = strings.TrimSpace(pair[0]), strings.TrimSpace(pair[1])
 	memLimit := resources.Limits[corev1.ResourceMemory]
 	memLimitByte := memLimit.Value()
 	if memLimitByte <= 0 {
+		// sidecar resources are only settled in the final pass, so until then a percentage may still be waiting for its limit
+		if finalPass && strings.HasSuffix(pair[1], "%") {
```
zxh326: Remove finalPass and drop buffer-size only when it is percentage-based and no memory limit is available.

For sidecar mode, resourcePercentages and the percentage-based buffer-size should be configured in the same mountPodPatch. The patch will be applied again when the sidecar is generated, allowing buffer-size to be calculated from the final memory limit.

Looka149: I removed `finalPass`, and also the log I added in the previous commit, to avoid confusion.

zxh326 (approved):

PR #1675, opened 2026-09-07
feat(dashboard): support sidecar batch binary upgrade

zwwhdls:
## What this PR does

Adds batch **binary upgrade** support for sidecar-mode JuiceFS clients, alongside the existing mount pod batch upgrade.

Unlike mount pods, sidecars cannot be recreated independently of their application pod, so the upgrade is done in-place: a canary job copies the new binary into the sidecar container, then the client is reloaded via `SIGHUP` without restarting the application pod.

## Implementation

**Upgrade flow** (`pkg/fuse/grace/sidecar_upgrade.go`)

1. Resolve the target image from the PVC/Secret settings and global config
2. Launch a canary job in the **system namespace** that `kubectl cp`s the new binary into the target sidecar
3. Upload the binary and send `SIGHUP` to the client process
4. Record the upgrade in a pod annotation and emit a pod `Event`

If the resolved target image is empty the sidecar is skipped (`POD-SKIP`) rather than treated as a failure.

**Refactor** — the shared upgrade orchestration was extracted from `grace.go` into a `GraceRunner` interface with two implementations, `pod_upgrade.go` (mount pod) and `sidecar_upgrade.go` (sidecar), so both kinds go through the same batching, status reporting and message protocol.

**Target discovery** (`pkg/config/batch_config.go`, `pkg/dashboard/services/pods/pod_service.go`) — `SelectSidecarUpgradeTargets` finds eligible sidecars by comparing the effective image against the resolved target image. `UpgradeTarget` gained a `ContainerName` field, and `Key()` returns `<pod>/<container>` for sidecars so status can be tracked per container.

**New API** — `GET /batch/upgrade/jobs/:jobName/diff` returns the current/target image of each sidecar container. Sidecars only change their binary, so the full mount pod config diff does not apply; `getUpgradeJob` returns an empty `diffs` list for sidecar jobs and skips the expensive PV/PVC/Secret lookup.

**Frontend** — the job detail page and batch upgrade modal now handle both kinds, showing an image diff for sidecars and the existing config diff for mount pods.

## Notable fixes included

- **Canary job namespace**: the job was created in the application namespace but waited for in the system namespace. Since `waitForJobCompleteQuiet` treats `IsNotFound` as success, the wait silently returned immediately and never actually waited for the binary copy to finish. Both now use the system namespace.
- **Upgrade progress resetting to pending**: the job detail page overwrote its status map with an empty `Map` on every refetch. Because the websocket `onClose` handler triggers a refetch, every pod dropped back to `pending` the moment the job finished. Status from the job config and from the websocket are now merged, with terminal states preserved.
- **Pause/resume needing several clicks**: `PUT /jobs/:jobName` returned as soon as the signal was delivered, so the follow-up refetch almost always read the pre-toggle status and the button appeared to do nothing. The handler now waits (bounded, 3s) for the status to actually flip, and the buttons are disabled while in flight.

## Testing

- `go build ./...`
- `go test ./pkg/config ./pkg/dashboard/... ./pkg/fuse/grace ./pkg/util/... ./pkg/juicefs/mount/builder ./cmd/dashboard`
- `tsc --noEmit` and `eslint` on the changed frontend files
- Manually verified against a live cluster: sidecar target discovery, batch execution, progress reporting, image diff, and pause/resume/stop

Note: `pkg/util` tests carry a `!darwin` build tag and are skipped locally, but run in CI.


pkg/fuse/grace/sidecar_upgrade.go
```diff
@@ -0,0 +1,429 @@
...
+	}
+
+	ownerReferences, err := resource.GetUpgradeJobOwnerReferences(ctx, r.client)
+	if err != nil {
+		return nil, err
+	}
+	job := builder.NewCanaryJobFromSpec(builder.CanaryJobSpec{
+		Name:               sidecarCanaryJobName(r.target),
+		Namespace:          config.Namespace,
+		Image:              r.targetImage,
```
zxh326: 要不要把 nodeName 传进来，避免架构问题

pkg/config/batch_config.go
```diff
@@ -62,12 +155,293 @@ const (
...
+		return "", false, nil
+	}
+	pvc, err := getPVC(pvcName, pod.Namespace)
+	if err != nil {
+		return "", false, err
+	}
+	if pvc == nil {
+		return "", false, nil
+	}
+	return GlobalConfig.GenMountPodPatch(JfsSetting{PVC: pvc, IsCe: isCe}, false, nil).Image, isCe, nil
```
zxh326: pod 所在的 node 信息需要带上，比如 acs 场景会使用专门的 nodeSelector 给 sidecar 应用不同的配置，不带的话应用不上

pkg/config/batch_config.go
```diff
@@ -62,12 +155,293 @@ const (
...
+
+// Key returns a key based on pod name and container name.
+func (t *UpgradeTarget) Key() string {
+	if t.Name == "" {
+		return ""
+	}
+	if t.ContainerName == "" {
+		return t.Name
+	}
+	return t.Name + "/" + t.ContainerName
```
zxh326: 加上 namespace？

pkg/config/batch_config.go
```diff
@@ -62,12 +155,293 @@ const (
...
+	if pod.Labels == nil || pod.Labels[common.InjectSidecarDone] != common.True {
+		return nil
+	}
+	containers := make([]corev1.Container, 0)
+	for _, container := range pod.Spec.Containers {
+		if isSidecarContainerName(container.Name) {
+			containers = append(containers, container)
+		}
+	}
+	if pod.Spec.RestartPolicy == corev1.RestartPolicyAlways {
```
zxh326: 这是为了处理原生 sidecar？

container.RestartPolicy == Always 才表示这个 init container 是原生 sidecar。

跟 pod.Spec.RestartPolicy 没有关系

pkg/fuse/grace/sidecar_upgrade.go
```diff
@@ -0,0 +1,429 @@
...
+		Command:            []string{"sh", "-c", "sleep 300"},
+		ServiceAccountName: common.UpgradeJobServiceAccountName(),
```
zxh326: 这个 env 应该没有传进来，要么就用 default sa 就行

pkg/fuse/grace/sidecar_upgrade.go
zwwhdls: serverless 没有 node 吧，不知道直接传 nodeName 会不会有问题

pkg/fuse/grace/sidecar_upgrade.go
zwwhdls: 之前是为了给 pod 用 kubectl 的权限，现在应该用不上了

pkg/fuse/grace/sidecar_upgrade.go
zxh326: serverless 的 nodename 都是 virtual-kubelet-xx

zwwhdls: 直接传 nodeName 不一定有用，可能需要跟 Pod 打一样的 annotation 或 label

pkg/config/batch_config.go
```diff
@@ -24,23 +24,116 @@ import (
...
+	}
+	if v, ok := raw["uniqueId"]; ok {
+		if err := json.Unmarshal(v, &bc.UniqueId); err != nil {
+			return err
+		}
+	}
+	if v, ok := raw["status"]; ok {
+		if err := json.Unmarshal(v, &bc.Status); err != nil {
+			return err
+		}
```
zxh326: 这里为啥需要这么复杂

pkg/fuse/grace/sidecar_upgrade.go
zwwhdls: 删了

pkg/fuse/grace/sidecar_upgrade.go
zwwhdls: 传了 nodeSelector 和 serverless 环境需要的 label/annotation

pkg/config/batch_config.go
```diff
@@ -181,15 +93,19 @@ type UpgradeTarget struct {
...
 
-// Key returns a key based on pod name and container name.
+// Key returns a key based on namespace, pod name and container name.
 func (t *UpgradeTarget) Key() string {
 	if t.Name == "" {
 		return ""
 	}
+	name := t.Name
+	if t.Namespace != "" {
+		name = t.Namespace + "/" + name
```
zxh326, line 103: 这里改了之后解析日志那里好像只有 podName。匹配不上？

zxh326: plugin 这边也需要适配，改回去也行，就是看看会不会有 不同 namespace 同名 pod 的情况

zxh326 (approved):

Issue #1653, opened 2026-08-19
[BUG] Node plugin exits with `fatal error: newosproc` when a mount stops answering


Looka149:
**What happened**:

One mount stopped answering. 12 to 14 hours later the node plugin exited on three nodes:

```
runtime: failed to create new OS thread (have 8099 already; errno=11)
runtime: may need to increase max user processes (ulimit -u)
fatal error: newosproc
```

We hit the pod cgroup `pids.max` (8192) before Go's 10000 limit, so the message is `newosproc`.

A syscall on an unresponsive mount never returns. `DoWithTimeout` times out the caller but leaves the goroutine behind. A goroutine blocked in a syscall keeps its OS thread, so every retry strands one more. Same mechanism as #1617.

https://github.com/juicedata/juicefs-csi-driver/blob/83b12a773f0101d59b9c33dc27cb6ec8f4cf532a/pkg/util/util.go#L329-L346

Goroutine dump just before the crash, by caller:

| caller | node A | node B | node C |
|---|---|---|---|
| `WaitUntilMountReady` | 7411 | 7864 | 7486 |
| `NodeGetVolumeStats` | 569 | 13 | 560 |
| `MkdirIfNotExist` | 84 | 240 | 12 |

Most of them are in `WaitUntilMountReady`, not in `NodeGetVolumeStats`.

`SupportFusePass` was off in our cluster, so the reconciler only retried and never aborted the FUSE connection. We do not think enabling it removes the problem:

- that path ends in `DeletePod`, and our `juicefs` process was in `D` state, which SIGKILL cannot end
- `NodeGetVolumeStats` runs on kubelet's timer and skips that branch either way

**What you expected to happen**:

A single Mount Pod should not be able to take down the node plugin.

**How to reproduce it (as minimally and precisely as possible)**:

In kind, with `SupportFusePass` off, `kill -STOP` the mount pod's `mount.juicefs` and `juicefs` processes (in production the mount stopped answering on its own). `stat` on the mount point then hangs and threads grow until the plugin exits.

**Anything else we need to know?**

**Environment**:
- JuiceFS CSI Driver version (which image tag did your CSI Driver use): v0.30.1, reproduced on v0.31.10
- Kubernetes version (e.g. `kubectl version`): v1.28.11, repro on v1.28.9
- Object storage (cloud provider and region):
- Metadata engine info (version, cloud provider managed or self maintained):
[...]

Issue #1617, opened 2026-07-13
[BUG] NodeGetVolumeStats may exhaust OS threads on an unhealthy FUSE mount


zxh326:
**What happened**:

When a JuiceFS/FUSE mount becomes unresponsive, os.Stat called by mount.PathExists may block indefinitely. DoWithTimeout returns a timeout, but the underlying goroutine and syscall continue running.

Repeated NodeGetVolumeStats calls then accumulate blocked OS threads until the process crashes:

```
runtime: program exceeds 10000-thread limit
fatal error: thread exhaustion

runtime stack:
runtime.throw({0x22319f5?, 0xc0248fbeb8?})
	/usr/local/go/src/runtime/panic.go:1073 +0x48 fp=0xc0248fbe78 sp=0xc0248fbe48 pc=0x474808
runtime.checkmcount()
	/usr/local/go/src/runtime/proc.go:903 +0x8e fp=0xc0248fbea0 sp=0xc0248fbe78 pc=0x43fa0e
runtime.mReserveID()
	/usr/local/go/src/runtime/proc.go:919 +0x2f fp=0xc0248fbec8 sp=0xc0248fbea0 pc=0x43fa4f
runtime.startm(0xc00009f408?, 0x1, 0x0)
	/usr/local/go/src/runtime/proc.go:3006 +0x112 fp=0xc0248fbf18 sp=0xc0248fbec8 pc=0x443212
runtime.wakep()
	/usr/local/go/src/runtime/proc.go:3161 +0xec fp=0xc0248fbf48 sp=0xc0248fbf18 pc=0x474acc
runtime.resetspinning()
	/usr/local/go/src/runtime/proc.go:3913 +0x3e fp=0xc0248fbf68 sp=0xc0248fbf48 pc=0x44575e
runtime.schedule()
	/usr/local/go/src/runtime/proc.go:4071 +0x10a fp=0xc0248fbfa0 sp=0xc0248fbf68 pc=0x445bea
runtime.mstart1()
	/usr/local/go/src/runtime/proc.go:1873 +0xcd fp=0xc0248fbfc8 sp=0xc0248fbfa0 pc=0x4418ad
runtime.mstart0()
	/usr/local/go/src/runtime/proc.go:1823 +0x76 fp=0xc0248fbff8 sp=0xc0248fbfc8 pc=0x4417b6
runtime.mstart()
[...]

PR #1692, opened 2026-09-16
fix: return stat errors from Exists instead of assuming the path exists

YuarenArt:
## Summary

- `util.Exists` treated every `stat` error except ENOENT as "exists", so on `ELOOP`, `ENOTDIR` or `ENAMETOOLONG` `MkdirIfNotExist` returned `nil` without creating anything and `createOrAddRef` / `mkrMp` carried on without a mount point directory. No error, no log line.
- `Exists` now returns `(bool, error)` with the semantics of `mount.PathExists`, which the driver already uses in `pkg/juicefs`: a corrupted mount point (ENOTCONN, ESTALE, EIO, EACCES) is `true, err`, anything else non-ENOENT is `false, err`.
- `MkdirIfNotExist` creates the directory only when the path is missing and returns other `stat` errors. A corrupted mount point keeps the previous contract and returns `nil`: it is still a mount point and the caller is about to recreate the mount pod for it, while `os.MkdirAll` on such a path would only fail with `EEXIST`.
- `ServeGfShutdown` in `pkg/fuse/grace` adjusted to the new signature.

Out of scope: the goroutine leak when `stat` blocks on an unresponsive FUSE mount (#1617, #1653) lives in `DoWithTimeout`; this change neither fixes nor worsens it.

## Test plan

- [x] `make verify`
- [x] `go test -gcflags=all=-l ./pkg/util/ -run 'TestExists|TestMkdirIfNotExist'` — existing, missing, `ELOOP`, `ENOTDIR`, and the corrupted case through a mode-0 directory (`EACCES`, skipped as root); the last row fails if the two checks in `MkdirIfNotExist` are swapped
- [x] `make test`

Related: #1617, #1653


YuarenArt:
@zxh326 gentle ping. Could you take a look when you have time? The CI is also waiting for approval to run. Thanks!

zxh326 (approved):

YuarenArt:
@zxh326 The `Test sanity` failure looks unrelated to this PR: master fails the same way since #1688 (https://github.com/juicedata/juicefs-csi-driver/actions/runs/34943430370). Could you confirm, and do you already know about it?

